Issue #45, opened 2023-03-09
Consider replacing ethers with viem


naps62:
This is a recent library. maybe worth exploring it here
https://twitter.com/wagmi_sh/status/1632814443972395010

naps62:
FYI, this has been reverted, but I'm not gonna re-open the issue, since this won't be possible in the foreseeable future.

Tauri relies on system webviews, which run specific JS targets (`es2021` I believe) which is incompatible with viem

jxom:
Out of curiosity, what features are missing in webviews that cannot be polyfilled? BigInts?

naps62:
@jxom to be fair, I didn't look into whether polyfills were an option or not (I figured throwing more unknowns into it would only cause further issues short-term)
but yes, it was about bigints. you can see the build logs here: https://github.com/iron-wallet/iron/actions/runs/4939277916/jobs/8829875517

jxom:
Ah right. That looks like a Vite build error which you can fix by setting the target in Vite to ES2021. Maybe that would fix it.

naps62:
I couldn't find clear answers about whether that was a good idea or not (example: https://github.com/tauri-apps/tauri-docs/issues/440)

the main difficulty here is that this will run on system's webviews. the safari version running on mac OS depends on the OS's version. I use linux myself, which makes it even harder to verify.

with that said, I'll probably get back to this once I have more time. I guess it was premature to say it "won't be possible in the foresseable future", so maybe after ETHGlobal Lisbon I can give a go at this again

PR #125, opened 2023-04-13
Replace Ethers with Viem

mirhamasala:
Fixes https://github.com/iron-wallet/iron/issues/45


gui/src/components/Txs.tsx
```diff
@@ -28,7 +29,7 @@ export function Txs() {
 
   return (
     <div className="bg-white shadow sm:rounded-lg">
-      <ul role="list" className="divide-y divide-gray-200">
+      <ul className="divide-y divide-gray-200">
```
mirhamasala, line 32: @naps62 A `ul` already has the role of a list. Am I missing something?

gui/src/components/Txs.tsx
```diff
@@ -88,18 +91,18 @@ function Receipt({ hash }: { hash: string }) {
   );
 }
 
-function Status({ status }: { status?: number }) {
+function Status({ status }: { status: TransactionReceipt["status"] }) {
```
mirhamasala, line 94: @naps62 TS was complaining here because `number` isn't compatible with the type `success | reverted` coming from `TrandactionReceipt` - Did I miss something?

gui/src/components/Txs.tsx
naps62: or maybe @zamith, our in-house accessibility advocate

gui/src/components/Txs.tsx
naps62: I'm not sure about this. on the surface. sounds like your code is correct, but I'm now wondering why it was working before

gui/src/components/Txs.tsx
mirhamasala: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role

gui/src/components/Txs.tsx
mirhamasala: Maybe because `status` is typed differently with Ethers?

Screenshot 2023-04-13 at 14 29 39

https://docs.ethers.org/v5/api/providers/types/#:~:text=receipt.status%20%E2%87%92%20number
vs. Viem

Screenshot 2023-04-13 at 14 30 30

https://github.com/wagmi-dev/viem/blob/0a148a983064a7183f030e63b9de0b3c9a0bc8d6/src/types/transaction.ts#L15

Either way, even if we don't go with Viem, should we be using the Ethers type for status instead of hardcoding `number`?

gui/src/components/Txs.tsx
```diff
@@ -1,12 +1,13 @@
 import { ArrowRightIcon } from "@heroicons/react/20/solid";
 import { listen } from "@tauri-apps/api/event";
 import classnames from "classnames";
-import { formatEther } from "ethers/lib/utils";
 import { useEffect } from "react";
 import useSWR from "swr";
 import truncateEthAddress from "truncate-eth-address";
+import { formatEther } from "viem";
+import type { TransactionReceipt } from "viem/src/types";
```
fubhy, line 8: You can't import from `viem/src` like that here. See `exports` in viem's package.json. This is causing the error that you reported in https://github.com/wagmi-dev/viem/issues/371

fubhy (changes requested):

gui/src/components/Txs.tsx
naps62: ohhh, they convert that to a string directly.
anyway. I think using `number` is perfectly fine while we stick with ethers
the spirit of typescript is that you don't need to specify the exact type all the time. If you imagine the codebase filled with `TransactionReceipt["xxx"]` instead of directly `string` or `number` or other basic types, it'd get a lot more confusing

changing to viem will cause typescript to show these errors, which (correctly) tells us that we need to change how we handle the value. so I'd say it's all working as intended

mirhamasala: Okay, I see.

mirhamasala: Are we not changing to Viem then? 😕

naps62: we are. I was answering your "even if we don't go with Viem" question

mirhamasala: @naps62 @gabrielpoca? Can we merge this, then? Or did I miss something that needs fixing still?

gabrielpoca: From what I understood yeah, this look ready, but we might want to do some manual testing before merging, I'll try to give it a go

gabrielpoca (approved):

PR #143, opened 2023-05-11
Increase the starting window's width

gabrielpoca:
So it doesn't look like this: 

Screenshot 2023-05-11 at 23 42 07

### DIFF
--- a/tauri/tauri.conf.json
+++ b/tauri/tauri.conf.json
@@ -48,7 +48,7 @@
         "fullscreen": false,
         "resizable": true,
         "title": "iron",
-        "width": 450,
+        "width": 1000,
         "height": 600
       }
     ]


naps62:
well, I noticed the same problem just moments ago.
but isnt' 1000 a bit excessive? at least with the current design

gabrielpoca:
@naps62 I think, ideally, we would remember the resize, that would be nice. As a default, I don't think it's too much but feel free to suggest something else. I may look into remembering the resize.

naps62:
yep. remembering the size would be the way to go
but btw, I don't think you need a plugin. this already has a `ContextInner` where we can store state, and that state is already backed up to a json config file

gabrielpoca:
@naps62 I handled that in another PR. What about the default value here? 900?

PR #144, opened 2023-05-13
Revamp interface with Material UI

gabrielpoca:
The whole thing started because I was building a button with the address that you could click to copy the address to the clipboard. I wanted a small tooltip saying "copy to clipboard" and "copied to clipboard" afterward. Doing it with tailwind and a headless component was too much work. Since we are still in the prototyping phase, I wanted more high-level components to build things quickly.

So, I changed the whole UI to use material UI. Most of the code is Lego work, combining components imported from material. This should make it easier for everyone that wants to contribute.

Most of this is just my replacing the existing UI elements with the material version, but I did take some time to refactor pieces of the code:
* extracted a bunch of functionality to hooks.
* flatten the components because a flat structure makes it simpler for contributors because they don't have to guess where to put components.
* revamped the transactions section.
* disabled autocorrect in some fields because it messed up the input text.
* other minor improvements.

And the cherry on top is that we get light and dark mode for free, and it adapts to the system configuration.


gui/src/types.ts
```diff
@@ -8,7 +8,7 @@ export const networkSchema = z.object({
         name: z.string().min(1),
         dev: z.boolean().default(false),
         http_url: z.string().min(1),
-        ws_url: z.string().optional(),
+        ws_url: z.string().nullable().optional(),
```
gabrielpoca, line 11: @naps62 have a look at this one. I had some issues because even though the field was optional, the value 'null' was not acceptable, only undefined.

gui/src/components/SettingsWallet.tsx
```diff
@@ -1,28 +1,39 @@
...
-export function WalletSettings() {
+export function SettingsWallet() {
   const { data: wallet, mutate } = useInvoke<Wallet>("get_wallet");
 
   const {
     register,
     handleSubmit,
     reset,
-    formState: { isValid, dirtyFields, errors },
+    formState: { isValid, isDirty, dirtyFields, errors },
```
naps62, line 29: @gabrielpoca I'm not sure if you saw the github issue linked a few lines below?

https://github.com/react-hook-form/react-hook-form/issues/3213

I was having problems with `isDirty` not working properly with nested objects. that's why `isDirtyAlt` is there instead.

gui/src/components/WagmiWrapper.tsx
```diff
@@ -36,7 +37,7 @@ export function WagmiWrapper({ children }: Props) {
     setClient(buildClient(network));
   }, [network]);
 
-  if (!client) return <></>;
+  if (!client) return null;
```
naps62, line 40: is this allowed? I remember seeing an error at some point when doing this

gui/src/types.ts
naps62: nice. I don't remember if I even thought about that before

naps62:
Love it. just need to see what's going on with that `isDirty` comment I left above.
but other than that let's merge this!

gui/src/components/WagmiWrapper.tsx
gabrielpoca: Yap, I think so, this is what I've always done. I also didn't see any error.

gui/src/components/SettingsWallet.tsx
gabrielpoca: yap, I used it by mistake 👍

naps62 (approved):

PR #145, opened 2023-05-13
Save window positions and sizes

gabrielpoca:
Installs `https://github.com/tauri-apps/tauri-plugin-window-state` and handles  `tauri::WindowEvent::CloseRequested` event to prevent the app from closing when the window is closed.

### DIFF
--- a/tauri/src/app.rs
+++ b/tauri/src/app.rs
@@ -133,21 +133,18 @@ impl IronApp {
 }
 
 fn on_window_event(event: GlobalWindowEvent) {
-    match event.event() {
-        WindowEvent::CloseRequested { api, .. } => {
-            let window = event.window();
-            let app = window.app_handle();
-            let _ = app.save_window_state(StateFlags::all());
+    if let WindowEvent::CloseRequested { api, .. } = event.event() {
+        let window = event.window();
+        let app = window.app_handle();
+        let _ = app.save_window_state(StateFlags::all());
 
-            #[cfg(target_os = "macos")]
-            app.hide().unwrap();
+        #[cfg(target_os = "macos")]
+        app.hide().unwrap();
 
-            #[cfg(not(target_os = "macos"))]
-            window.close().unwrap();
+        #[cfg(not(target_os = "macos"))]
+        window.close().unwrap();
 
-            api.prevent_close();
-        }
-        _ => {}
+        api.prevent_close();
     }
 }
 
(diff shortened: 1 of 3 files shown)


tauri/src/app.rs
```diff
@@ -92,10 +95,11 @@ impl IronApp {
     }
 
     pub fn run(&mut self) {
-        self.app.take().unwrap().run(|_, event| {
-            if let tauri::RunEvent::ExitRequested { api, .. } = event {
+        self.app.take().unwrap().run(|_, event| match event {
+            tauri::RunEvent::ExitRequested { api, .. } => {
                 api.prevent_exit();
             }
+            _ => {}
```
naps62, line 102: why this change? as you can see, clippy is already asking to change it back

gabrielpoca: oh, sorry, I was going to handle the event here

tauri/src/app.rs
```diff
@@ -129,6 +132,22 @@ impl IronApp {
...
+        let app = window.app_handle();
+        let _ = app.save_window_state(StateFlags::all());
+
+        #[cfg(target_os = "macos")]
+        app.hide().unwrap();
+
+        #[cfg(not(target_os = "macos"))]
+        window.close().unwrap();
+
+        api.prevent_close();
```
naps62, line 147: If I understood correctly, this keeps the UI process running the bg as well, presumably making it load faster once we open the window again?

If that's the case, I'm fine with the change, but can you first open an issue for a future task (maybe a `good-first-issue` thing), to add a settings option to switch this. Some people want will want to run this purely as a daemon, and not have the UI spam RPC requests they won't use

gabrielpoca: Ok, so this was a big rabbit hole.

In macOS, you have both the system tray/menubar and the dock. Most apps don't use the system tray, and when you close a window, it doesn't close; it is hidden (minimizing is a different thing). When you then click the icon on the dock, the window opens again.

With this information, I disabled the system tray on macOS. Some apps have a setting to change this behavior, for instance:
Screenshot 2023-05-16 at 11 44 39
Screenshot 2023-05-16 at 11 44 46

These are Telegram and TickTick. I made the executive decision to stick with the dock for now because it's the typical approach, I think 😅

Either way, when taking that route, we [don't yet have a listener for dock events](https://github.com/tauri-apps/tauri/issues/3084), but using the `prevent_close` solves it. So I changed that code to only run on macOS.

naps62 (approved):

PR #147, opened 2023-05-15
Clicking on addresses, hashes and values copies to clipboard

gabrielpoca:
A small change, a big quality of life improvement. Plus, the component `CopyToClipboard` can be easily used anywhere. You don't need to specify the text you want to copy to the clipboard.

### DIFF
--- a/gui/src/components/Balances.tsx
+++ b/gui/src/components/Balances.tsx
@@ -9,6 +9,7 @@ import { useAccount } from "../hooks";
 import { useInvoke } from "../hooks/tauri";
 import { useRefreshTransactions } from "../hooks/useRefreshTransactions";
 import { Address } from "../types";
+import { CopyToClipboard } from "./CopyToClipboard";
 import Panel from "./Panel";
 
 export function Balances() {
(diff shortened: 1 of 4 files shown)


naps62:
One thing is annoying me
when I hover something, it takes some miliseconds for the tooltip to show up (which is good, less intrusive). But after that first one, moving the mouse throw the window shows the tooltip immediatelly on other elements. On the Transactions tab, where there's a bunch of these, it gets pretty annoying

Plus, I feel that with the `cursor: pointer` and the hover background color, it's already noticeable enough that this is clickable. maybe we could go without the tooltip, and just show it when it is clicked?

gabrielpoca:
@naps62 Yap, the delay was on purpose, so you only see it if you hover for a bit. I fixed the issues you mentioned and increased the delay time a bit. This should be enough so you never see those tooltips, but let me know what you think. I'm ok with removing them, but hovering on elements to see the tooltip is an action I frequently do; that's why I built it.

naps62 (approved):

PR #156, opened 2023-05-16
Initial support for keyboard navigation

gabrielpoca:
This change takes the first steps toward keyboard navigation:
* Use routes to change tabs. Using routes instead of a component's internal state means we can change tabs from anywhere. This logic will apply to everything.
* Create menubar items. The menubar allows people to set local or global shortcuts in the app. At least on macOS, you can easily map shortcuts to items in the menubar.
* Defining initial in-app shortcuts to change tabs. We can replace these later, this is just an initial step.

### DIFF
--- a/gui/src/components/NestedRoutes.tsx
+++ b/gui/src/components/NestedRoutes.tsx
@@ -1,0 +1,24 @@
+import React from "react";
+import { ReactNode } from "react";
+import { Router, useLocation, useRouter } from "wouter";
+
+export function NestedRoutes({
+  children,
+  base,
+}: {
+  children: ReactNode;
+  base: string;
+}) {
+  const router = useRouter();
+  const [parentLocation] = useLocation();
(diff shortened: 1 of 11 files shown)


gui/src/components/NestedRoutes.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { ReactNode } from "react";
+import { Router, useLocation, useRouter } from "wouter";
```
naps62, line 3: weren't you dropping wouter in favor of something else?

naps62 (approved):

gui/src/components/NestedRoutes.tsx
gabrielpoca: Yeah, I decided to keep it. React Router keeps breaking its API, so we might as well stick with this.

PR #158, opened 2023-05-17
Dialog windows (behind a feature flag, since UI/UX is still WIP)

naps62:
This is a big one, because I had to change a few things, but I think it sets some good standards for the future.

It mainly creates the infrastructure for being able to trigger arbitrary popup dialogs from the backend.
These dialogs are suposed to return a success or failure message back (e.g.: to accept or reject a transaction; to allow/forbid some external connection; etc)

I'm thinking of these dialogs as if they were an actual function of type `pub async fn dialog(msg) -> Result<Accept, Reject>`.

image

## Notable changes

- Moved the transaction building logic into a `SendTransaction` struct, as a sort of builder object. This is to clean up the RPC method, and to later make it easier to expand on that logic
- the RPC method first calls the popup, waits for its response. If it's accepted, proceeds to build the `SendTransaction` struct
- `dialog.rs` encapsulates all dialog-related logic
- the entire path for creating a dialog is:
  - `dialogs::open(...)`, sends a msg to the Tauri event loop
  - a oneshot channel is created and stored globally. this channel will later receive the response from the dialog
  - once tauri picks up the event, it calls `dialogs::open_with_handle(...)`. this indirection is necessary because we can't store a global app_handle
  - the dialog opens, and we expect it to call `dialog_finish` with a result object
  - `dialog_finish` will send this object to the pending oneshot channel, and close the dialog

There's already a barebones example of a dialog for transaction confirmation. This is pending design, which I'll implement in a future pull request.

## Future work

- Make the dialog presentable
- Add a settings option to toggle the need for dialogs (in livenets they'll always show, but we can opt-out when using anvil)
- remove the feature flag to make the feature available by default


gui/src/components/TxReviewDialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+  to: string;
+  value: string;
+}
+type Props = {
+  id: number;
+  payload: TxRequest;
+};
+
+export function TxReviewDialog({ id, payload }: Props) {
+  console.log(payload);
```
gabrielpoca, line 18: bump

gui/src/hooks/useDialog.ts
```diff
@@ -0,0 +1,22 @@
...
+export function useDialog<T>(id: number) {
+  const { data } = useInvoke<T>("dialog_get_payload", { id });
+
+  const accept = useCallback(
+    (payload: unknown) =>
+      invoke("dialog_finish", { id, result: { Ok: payload } }),
+    [id]
+  );
+
+  const reject = useCallback(
```
gabrielpoca, line 15: Can you remove the `useCallback`? These really just for optimizations.

naps62: I was researching this (after noticing you didn't use them).
one of the rules-of-thumb I read about this was: use `useCallback` if you pass the callbacks down to a child

is that not applicable here? or rather, is there any downside to using it here? feels like it could eventually become one of those hard to track bugs

gabrielpoca: I didn't find the talk, but here's a [blog post about it](https://kentcdodds.com/blog/usememo-and-usecallback).

naps62: fixed

gabrielpoca (approved):

PR #159, opened 2023-05-17
Accordion

naps62:
Puts each network inside an accordion item by default

Adding a new network creates a new item expanded by default

image

### DIFF
--- a/gui/src/components/SettingsNetwork.tsx
+++ b/gui/src/components/SettingsNetwork.tsx
@@ -75,7 +75,6 @@ export function SettingsNetwork() {
       {fields.map((field, index) => {
         //        const item = networks[index];
         const err = (errors.networks && errors.networks[index]) || {};
-        console.log(field);
         return (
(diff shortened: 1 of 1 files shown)


gui/src/components/SettingsNetwork.tsx
```diff
@@ -64,98 +72,104 @@ export function SettingsNetwork() {
...
-        {fields.map((field, index) => {
-          //        const item = networks[index];
-          const err = (errors.networks && errors.networks[index]) || {};
-          return (
-            <Stack spacing={2} alignItems="flex-start" key={field.id}>
-              <Stack spacing={2} direction="row">
+      {fields.map((field, index) => {
+        //        const item = networks[index];
+        const err = (errors.networks && errors.networks[index]) || {};
+        console.log(field);
```
gabrielpoca, line 78: bump

naps62: :( keep forgetting these

gabrielpoca (approved):

PR #160, opened 2023-05-18
Add explorer url to the network settings

gabrielpoca:
This will allow users to automatically open certain information in their explorer of choice.


gui/src/components/SettingsNetwork.tsx
```diff
@@ -187,6 +199,14 @@ export function SettingsNetwork() {
...
         </Button>
       </Stack>
+      <Button
+        variant="outlined"
+        color="warning"
+        size="medium"
+        onClick={onReset}
+      >
+        Reset Networks
+      </Button>
```
naps62, line 209: What's the need for this?
also, sounds rather dangerous to have a 1-click reset of important settings

gabrielpoca: It's for when I mess up the networks and want the defaults back. I can add a confirmation modal maybe?

naps62: I think that would help, yes

PR #163, opened 2023-05-18
Support for multiple wallets

naps62:
This was a hard one. here's a gist of the changes: 

image

## Features

- We now have a `Vec<Wallet>` instead of a single `Wallet`
- The settings now work similarly to the network ones, with an accordion
- for each wallet, instead of a list of radios, we currently choose a starting `index` and a `count` instead. This also means we no longer have a hardcoded list of 5 addresses for each wallet
- there's two wallet-related dropdowns now. one to choose the wallet, and another to choose the address within that wallet

## Code details

- There was too much coupling inside the `Context` that made it hard to work with. So a lot changed in that regard. I created `crate::wallet` and `crate::peers` which have their own global state, instead of everything sharing the same global Context. This is more in-line with Rust best practices. having a single state object means we need to lock the entire world for every access. Instead, we should be have more fine-grained locking
- I also used `RwLock`, which should allow concurrent reads (both from the UI and from connected tabs). It's not that this was a problem so far, but seems like the right abstraction anyway, and it's the same effort as using a regular `Mutex`
- There's a `GlobalState` trait to define how global state objects should behave: `init`, `read` and `write`. It's a simple API, but I find it useful to have it explicit there already
- both `peers` and `wallet` now own their own commands. The idea is that each module is as isolated as possible. Eventually we may need to extract some to sub-crates (e.g.: to speed up compilation)
- The context still exist, but I'll probably phase it out in favor of more fine grained objects
- Even though there's global state, each of those also behaves perfectly well as a regular object, so testing shouldn't be a problem if we need to
- The only point of coupling between those modules is that `Wallets` calls `Peers` to send broadcast messages. Ideally this can be done with a channel instead of a direct call. But that will be done later if necessary
- commands now are prefixed with the module name. this makes it more clear in the UI code, in my opinion


naps62:
I still need to fix some of the TODOs in here

gui/src/types.ts
```diff
@@ -23,15 +23,26 @@ export const networkSchema = z.object({
...
+  dev: z.boolean().default(false),
   mnemonic: z.string().regex(/^(\w+\s){11}\w+$/, {
     message: "Must be a 12-word phrase",
   }),
   derivationPath: z.string().regex(/^m\/(\d+'?\/)+\d+$/, {
     message: "invalid path format",
   }),
   idx: z.number().int().min(0).max(4),
+  count: z.number().int().min(1),
+  currentKey: z.string().optional(),
```
gabrielpoca, line 39: `currentKey` is an index? address? What is it?

gui/src/components/Connections.tsx
```diff
@@ -17,7 +17,7 @@ interface Connection {
 
 export function Connections() {
   const { data: connections, mutate } =
-    useInvoke<Connection[]>("get_connections");
+    useInvoke<Connection[]>("peers_get_all");
```
gabrielpoca, line 20: It feels like we are using `peer` and `connection` interchangeably. Don't we want some consistency here?

gui/src/types.ts
naps62: it's the index of which address is currently selected in this wallet

it's not part of the form, though. just needs to exist on the infered Wallet type

gui/src/components/Connections.tsx
naps62: true. I renamed some things while working on this, but didn't do it everywhere

do you have a preference? I think Peer is better here, since it conveys better that it's an external client connected to us

gui/src/types.ts
naps62: to be honest don't like either of those, although the current one is also not great

my thinking is that `currentIndex` makes it sound like an array with numeric indexes, and `currentAccount` introduces yet another ambiguous word (wallet, address, account, it becomes very confusing)

the thinking behind key is that it's suposed to be agnostic about what that means inside the wallet, as if the wallet was a HashMap<key, Address>

gui/src/components/Connections.tsx
gabrielpoca: But it's a client (browser tab) server (our backend) relationship. But either way, you choose.

gui/src/types.ts
gabrielpoca: `currentPath`? Path is a normal word in derivation paths.

gui/src/components/Connections.tsx
naps62: my problem with "Connection" btw, is that we have many types of connections already (e.g.: to an ETH provider).
perhaps "Client"?

gabrielpoca: "Connected Tabs" myabe? or "Connected Websites"?

naps62: it's neither of those. iframes create their own connections as well

gabrielpoca: 🙈

gui/src/types.ts
naps62: renamed to `currentPath`

gui/src/components/Connections.tsx
naps62: renamed to peers (not that I love it, but because we can't keep this thread forever xD)

naps62:
@gabrielpoca now ready for review. or you can just test it, not sure if you want to review the Rust side

gabrielpoca (approved):

PR #182, opened 2023-05-22
Command bar

gabrielpoca:
This change introduces a command bar with actions to change networks and color schemes.
To use this command bar for everything, we must also write the application using more providers, as I did by introducing those two new providers. Each provider can register its actions that can be used by the command bar or the other elements on the screen.


tauri/src/networks/mod.rs
```diff
@@ -65,24 +64,46 @@ impl Networks {
...
-        self.save().unwrap();
+    pub fn reset_networks(&mut self) -> Result<()> {
+        self.do_set_networks(Network::default())
     }
 
     pub fn get_current_provider(&self) -> Provider<Http> {
         self.get_current_network().get_provider()
     }
 
+    fn broadcast_network_changed(&self) -> Result<()> {
```
naps62, line 88: I don't like this name, because it's mixing nomenclature with the `broadcast_chain_changed` which is part of EIP-1193
this is doing 2 different things. that EIP-1193 part, plus notifying our own frontend

how afte `after_network_changed` or `on_network_changed`?

gabrielpoca: 👍 sure. I didn't think too much about it because it's a private method

tauri/src/networks/mod.rs
```diff
@@ -65,24 +64,46 @@ impl Networks {
...
+        self.window_snd.send(app::Notify::NetworkChanged.into())?;
+
+        Ok(())
+    }
+
+    fn do_set_networks(&mut self, networks: HashMap<String, Network>) -> Result<()> {
+        self.networks = networks;
+
+        if !self.networks.contains_key(&self.current) {
+            let name = self.networks.values().next().unwrap().name.clone();
```
naps62, line 99: this is not deterministic. HashMaps don't have a fixed order, as they would in JS, so this essentially means the current network will be a random one, and may not even be the same between runs, or between computers.

That's also one of the reasons why the `set_networks` state is a `Vec` and not a `HashMap` as well.

my suggestion: have this function receive a `Vec<Network>` instead (which will also simplify the `set_networks` call), and change `reset_networks` to create a vec with a hardcoded order`

then you can use `networks[0].name` as the fallback one

gui/src/hooks/useTheme.tsx
```diff
@@ -1,37 +1,7 @@
...
-  const final =
-    data?.darkMode == "dark" || (data?.darkMode == "auto" && prefersDarkMode);
-
-  const theme = useMemo(
-    () => createTheme(getDesignTokens(final ? "dark" : "light")),
-    [final]
-  );
-
-  return theme;
+  return useContext(ThemeContext);
```
naps62, line 6: is this file even needed now? 😅

gui/src/components/SettingsNetwork.tsx
```diff
@@ -64,15 +62,13 @@ export function SettingsNetwork() {
   });
 
   const onSubmit = async (data: { networks?: Network[] }) => {
-    await invoke("set_networks", { networks: data.networks });
+    if (data.networks) await setNetworks(data.networks);
```
naps62, line 65: shouldn't this guard cover the `reset` as well?

(I'm not sure when this would actually happen)

gui/src/components/ProviderTheme.tsx
```diff
@@ -0,0 +1,75 @@
...
+        }
+      : {
+          background: {},
+        }),
+  },
+});
+
+export const ThemeContext = createContext<Theme | null>(null);
+
+export function ProviderTheme({ children }: { children: ReactNode }) {
```
naps62, line 35: maybe I'm confused but shouldn't the name be `ThemeProvider` instead?

gui/src/components/ProviderNetworks.tsx
```diff
@@ -0,0 +1,73 @@
...
+  setNetwork: (newNetwork: string) => Promise<unknown>;
+  setNetworks: (newNetworks: Network[]) => Promise<unknown>;
+  resetNetworks: () => Promise<Network[]>;
+}
+
+export const NetworksContext = createContext<Value>({} as Value);
+
+const actionId = "network";
+
+export function ProviderNetworks({ children }: { children: ReactNode }) {
```
naps62, line 21: same here. I'm confused about this naming

tauri/src/networks/mod.rs
gabrielpoca: > so this essentially means the current network will be a random one, and may not even be the same between runs, or between computers.

I need some clarification. This code is only executed when you update the list of networks and the new list doesn't include the currently selected network. Sure, it picks a random one, but that one will be persisted, and the subsequent runs will always use that until the user changes it. Does it matter that the selection is random? I'm happy to change it, but it feels a bit irrelevant. But feel free just to say that you think it's important; there's no need to make it a big thing.

gui/src/components/SettingsNetwork.tsx
gabrielpoca: oh, yeah 👍

gui/src/components/ProviderTheme.tsx
gabrielpoca: I like to use prefixes instead so all providers live side-by-side in the file system. Do you think that's weird?

gui/src/components/ProviderNetworks.tsx
```diff
@@ -0,0 +1,73 @@
...
+  useRefreshNetwork(mutateNetwork);
+
+  useRegisterActions(
+    [
+      {
+        id: actionId,
+        name: "Change network",
+      },
+      ...(networks || []).map((network) => ({
+        id: `${actionId}{network.name}`,
```
naps62, line 59: this feels a bit weird without a separator. can we add a `/` in between?

gui/src/components/ProviderTheme.tsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  useRegisterActions([
+    {
+      id: themeModeId,
+      name: "Change theme mode",
+      section: "",
+    },
+    ...["auto", "dark", "light"].map((mode) => ({
+      id: `${themeModeId}${mode}`,
```
naps62, line 55: same here with the `/`

gui/src/components/ProviderTheme.tsx
naps62: it's fine with me, now that I know the reason. I honestly thought it was by mistake 😅 

but since you mentioned it: wouldn't a more appropriate solution be to have a `Providers` directory instead of a flat `components`?

(not requesting changes here, consider this one solved)

gui/src/components/ProviderNetworks.tsx
gabrielpoca: Does it matter? It's just a random id, but I can add it

gui/src/components/ProviderNetworks.tsx
```diff
@@ -0,0 +1,73 @@
...
+      mutateNetworks();
+    },
+    resetNetworks: async () => {
+      const networks: Network[] = await invoke("networks_reset");
+
+      mutateNetworks();
+
+      return networks;
+    },
+  };
```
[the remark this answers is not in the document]
gabrielpoca: 👍 honestly, the names are very bad :/ Having an "s" to distinguish things is no good. I'll take option 1 because option 2 is not solving the problem since destructing is a common pattern.

gui/src/components/ProviderNetworks.tsx
naps62: probably not. I initially understood that these ids were used by us somehow, but I think I got it wrong

PR #184, opened 2023-05-23
Adds a windows-compatible regex for foundry ABI paths

naps62:


naps62:
This was tested manually with @ZePedroResende . seems to be working

PR #202, opened 2023-05-29
Ability to fetch ERC20 balances from alchemy.com

CondeGil:
image


tauri/src/db/mod.rs
```diff
@@ -45,6 +45,26 @@ impl DB {
         Ok(db)
     }
 
+    pub async fn save_balances(&self, balances: AlchemyResponse, chain_id: u32) -> Result<()> {
+        let mut conn = self.tx().await?;
+
+        for balance in balances.token_balances {
+            if !balance.token_balance.is_zero() {
```
CondeGil, line 52: I added this check here, but this seems to me a bad idea. I think this check should be made on the front end side, here we should always add to the DB.

gui/src/components/SettingsNetwork.tsx
```diff
@@ -134,6 +135,15 @@ export function SettingsNetwork() {
                     helperText={err.decimals?.message?.toString()}
                   />
                 </Stack>
+                {field.chain_id !== 31337 && ( // Only show this field in live networks
+                  <TextField
+                    label="Alchemy URL (with API key)"
```
CondeGil, line 140: add a placeholder would be better?

tauri/Cargo.toml
```diff
@@ -45,6 +49,7 @@ futures = "0.3.28"
 regex = "1.8.1"
 anyhow = "1.0.71"
 secrets = "1.2.0"
+reqwest = { version = "0.11", features = ["json"] }
```
naps62, line 52: can you add the `tokio-rustls` feature here?
I think that should get rid of the extra dependency that showed up on Cargo.lock, and it's also more consistent to avoid using two different TLS implementations

tauri/src/types/alchemy_response.rs
```diff
@@ -0,0 +1,16 @@
...
+    pub address: Address,
+    pub token_balances: Vec<TokenBalance>,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TokenBalance {
+    pub contract_address: Address,
+    pub token_balance: U256,
+}
```
naps62, line 16: I think this new type should live on the same module as the listener, since it seems to be a more scoped thing
this types module is intended for broad things (e.g.: types that will be used in many places, or that need to be replicated on the UI as well)

tauri/src/db/mod.rs
naps62: well, the check here will just be wrong.
e.g.: if you have 10 USDC and then transfer them out, the 2nd transaction won't go through here, and the UI will still how 10 USDC in your balance

I would store everything here. the decision of whether to show tokens with zero balances is a different feature (probably a combination of having an option in the UI, and also some automatic filter to hide the spam coins)

tauri/src/live_networks_listener/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub fn stop(&mut self) {}
+}
+
+async fn process(ws_url: Url, db: DB, chain_id: u32) -> Result<()> {
+    let address = Wallets::read()
+        .await
+        .get_current_wallet()
+        .get_current_address()
+        .await;
+    let address = json!(address); // didnt find a better way to convert it to &str
```
naps62, line 46: currently there's no way, but there should. easiest way is to just change the type:
```
pub struct ChecksummedAddress(Address);
```

into

```
pub struct ChecksummedAddress(pub Address);
```


then you can do `address.0`

---

but I think a better approach would be to have the listener receive the address when it starts (and we just restart to a new listener if the current account changes). that's a deeper change though

tauri/Cargo.toml
CondeGil: Im not seeing that feature here: https://docs.rs/reqwest/latest/reqwest/
Where can i find more about that?

gui/src/components/SettingsNetwork.tsx
naps62: what do you mean?

CondeGil:
```
<TextField
    label="Alchemy URL (with API key)"
    placeholder="https://eth-mainnet.g.alchemy.com/v2/XXXXXXXXXXXXXXXXXXXXXXXXXXX"
    ...
/>
```

naps62: ahh. makes sense yeah

gui/src/types.ts
```diff
@@ -4,6 +4,8 @@ export const generalSettingsSchema = z.object({
   darkMode: z.enum(["auto", "dark", "light"]),
   abiWatch: z.boolean().default(false),
   abiWatchPath: z.string().optional(),
+  hideEmptyTokens: z.boolean().default(true),
```
naps62, line 7: this shouldn't need a default on the frontend. that comes from the backend

gui/src/components/Balances.tsx
```diff
@@ -18,22 +18,30 @@ export function Balances() {
...
-            key={contract}
-            {...{
-              contract,
-              balance: BigNumber.from(balance),
-            }}
-          />
-        ))}
+        {(balances || [])
+          .filter(([, balance]) =>
+            settings?.hideEmptyTokens ? !BigNumber.from(balance).isZero() : true
```
naps62, line 31: I moved to viem in a recent PR, so you'll need to change this to use `BigInt` instead of `BigNumber`

gui/src/components/Balances.tsx
```diff
@@ -18,22 +18,30 @@ export function Balances() {
...
+          )
+          .map(([contract, balance]) => (
+            <BalanceERC20
+              key={contract}
+              {...{
+                contract,
+                balance: BigNumber.from(balance),
+              }}
+            />
+          ))}
```
naps62, line 44: can you move this chian to a separate variable? to keep the markup here cleaner

gui/src/types.ts
CondeGil: Yeah, noticed that as well. The reason i put a default was because the other fields have one as well. Look at abiWatch for instance

naps62: yeah, that one is wrong too. my bad 😅

tauri/src/db/mod.rs
```diff
@@ -46,6 +47,24 @@ impl DB {
         Ok(db)
     }
 
+    pub async fn save_balances(&self, balances: AlchemyResponse, chain_id: u32) -> Result<()> {
```
naps62, line 50:
```suggestion
    // TODO: Change this to an Into<T> of some kind, so we don't depend directly on AlchemyResponse here
    pub async fn save_balances(&self, balances: AlchemyResponse, chain_id: u32) -> Result<()> {
```

tauri/src/alchemy/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub contract_address: Address,
+    pub token_balance: U256,
+}
+
+impl Alchemy {
+    pub fn new(db: DB) -> Self {
+        Self { db }
+    }
+
+    async fn fetch_balances(&self) -> Result<()> {
```
naps62, line 58: this should get the chain_id and address as parameter instead of fetching the current one

the idea is to decouple those concepts. Alchemy doesn't need to care what the "current" address is, since we may decide to fetch others in the background

tauri/src/alchemy/global.rs
```diff
@@ -0,0 +1,27 @@
...
+
+static ALCHEMY: OnceCell<RwLock<Alchemy>> = OnceCell::new();
+
+#[async_trait]
+impl GlobalState for Alchemy {
+    type Initializer = DB;
+
+    async fn init(db: Self::Initializer) {
+        let instance = Alchemy::new(db);
+        instance.fetch_balances().await.unwrap();
```
naps62, line 16: I don't think we should call this here (especially because it will receive arguments)

probably the real solution is that the UI itself will send an event asking for the balances to be refreshed (e.g.: if the GUI window isn't open, we don't even need to bother updating)

tauri/src/alchemy/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+                    "method": "alchemy_getTokenBalances",
+                    "params": [address, "erc20"]
+                }))
+                .send()
+                .await?
+                .json()
+                .await?;
+
+            let res = (&res["result"]).clone();
+            let res: AlchemyResponse = serde_json::from_value(res).unwrap();
```
naps62, line 81: these `unwrap` needs to become a `?`, which probably means adding a new error type

naps62 (commented):
almost there 💪🏼

tauri/src/alchemy/commands.rs
```diff
@@ -0,0 +1,9 @@
+use super::{Alchemy, Result};
+use crate::types::{ChecksummedAddress, GlobalState};
+
+/// call to fetch balances
```
naps62, line 4:
```suggestion
/// call to fetch balances
/// TODO: this should also trigger a frontend event to refresh the balances if the request succeeds
```

gui/src/components/Balances.tsx
```diff
@@ -1,22 +1,38 @@
...
+    (async () => {
+      if (currentNetwork && address) {
+        await invoke("alchemy_fetch_balances", {
+          chainId: currentNetwork.chain_id,
+          address,
+        });
+        mutate();
+      }
+    })();
+  }, [address, currentNetwork, mutate]);
```
naps62, line 34: let's keep this file out of this PR please.
I'm also not sure if this is the best strategy to fetch, since it will trigger every time the UI loads

naps62 (approved):

PR #215, opened 2023-05-31
hide empty tokens in details ui

CondeGil:
Add a checkbox in settings to allow users to decide if they want to see empty tokens in the details ui page or not.
image
image

### DIFF
--- a/gui/src/components/Balances.tsx
+++ b/gui/src/components/Balances.tsx
@@ -6,7 +6,7 @@ import { useBalance, useContractRead } from "wagmi";
 import { useAccount } from "../hooks";
 import { useInvoke } from "../hooks/tauri";
 import { useRefreshTransactions } from "../hooks/useRefreshTransactions";
-import { Address } from "../types";
+import { Address, GeneralSettings } from "../types";
 import { CopyToClipboard } from "./CopyToClipboard";
 import Panel from "./Panel";
 
@@ -16,6 +16,12 @@ export function Balances() {
     "db_get_erc20_balances",
     { address }
   );
+  const { data: settings } = useInvoke<GeneralSettings>("settings_get");
+
+  const reorderedBalances = (balances || [])
+    .map<[`0x${string}`, bigint]>(([c, b]) => [c, BigInt(b)]) // is it possible to get the proper type from backend directly?
+    .filter(([, balance]) => (settings?.hideEmptyTokens ? !!balance : true))
+    .sort(([, a], [, b]) => (b > a ? 1 : b < a ? -1 : 0));
 
   useRefreshTransactions(mutate);
 
@@ -23,13 +29,11 @@ export function Balances() {
     <Panel>
       <Stack>
         {address && <BalanceETH address={address} />}
-        {(balances || []).map(([contract, balance]) => (
+        {reorderedBalances.map(([contract, balance]) => (
           <BalanceERC20
             key={contract}
-            {...{
-              contract,
-              balance: BigInt(balance),
-            }}
+            contract={contract}
+            balance={BigInt(balance)}
           />
         ))}
       </Stack>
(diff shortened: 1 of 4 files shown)


gui/src/components/Balances.tsx
```diff
@@ -16,20 +16,24 @@ export function Balances() {
     "db_get_erc20_balances",
     { address }
   );
+  const { data: settings } = useInvoke<GeneralSettings>("settings_get");
+
+  const reorderedBalances = (balances || [])
+    .map<[`0x${string}`, bigint]>(([c, b]) => [c, BigInt(b)]) // is it possible to get the proper type from backend directly?
```
gabrielpoca, line 22: It's weird that you must type it because it works on line 32, or do you see an error there too?

gui/src/components/Balances.tsx
```diff
@@ -16,20 +16,24 @@ export function Balances() {
...
+    .filter(([, balance]) => (settings?.hideEmptyTokens ? !!balance : true))
+    .sort(([, a], [, b]) => (b > a ? 1 : b < a ? -1 : 0));
```
gabrielpoca, line 24: What do you think of using the `sortBy` from lodash? I feel like it's more readable since you can just return the balance and don't need to write these conditions

gui/src/components/Balances.tsx
naps62: btw, for the `0x{string}` you already have an `Address` type

gui/src/components/Balances.tsx
naps62: I actually think we shouldn't have this `sort` right now
firstly because sorting by balance is very arbitrary, and will mean the list changes order with every transaction

secondly because we probably want to have some other order defined by the backend, e.g. by the order in which they were first detected (which is what metamask does)

gui/src/components/Balances.tsx
CondeGil: The types gets wrong without that cast (`(bigint | '0x${string}')[][]`), but here its not necessary to be fair. That's a residue from an old implementation. Ill remove it.

gui/src/components/Balances.tsx
CondeGil: yeah, i get that. my idea was just to bring the tokens with balance (!== 0) to the top in case you're not filtering those

naps62 (approved):

tauri/src/settings/mod.rs
```diff
@@ -80,6 +80,7 @@ pub struct SerializedSettings {
 
     pub abi_watch: bool,
     pub abi_watch_path: Option<String>,
+    pub hide_empty_tokens: bool,
```
naps62, line 83:
```suggestion
    #[serde(default = true)]
    pub hide_empty_tokens: bool,
```

this needs to be done so that existing installations still work when deserializing an existing settings file

gui/src/components/Balances.tsx
```diff
@@ -16,20 +16,23 @@ export function Balances() {
...
-        {(balances || []).map(([contract, balance]) => (
+        {reorderedBalances.map(([contract, balance]) => (
           <BalanceERC20
             key={contract}
-            {...{
-              contract,
-              balance: BigInt(balance),
-            }}
+            contract={contract}
+            balance={BigInt(balance)}
```
naps62, line 35: this can now be just `balance={balance}` right?

gui/src/components/Balances.tsx
```diff
@@ -16,20 +16,23 @@ export function Balances() {
...
+    .filter(([, balance]) => (settings?.hideEmptyTokens ? !!balance : true));
 
   useRefreshTransactions(mutate);
 
   return (
     <Panel>
       <Stack>
         {address && <BalanceETH address={address} />}
-        {(balances || []).map(([contract, balance]) => (
+        {reorderedBalances.map(([contract, balance]) => (
```
naps62, line 31: now it's no reordering, so maybe just `filteredBalances`?

gui/src/components/Balances.tsx
CondeGil: yap, my bad

naps62 (approved):

PR #216, opened 2023-05-31
add text field for alchemy api key in ui settings

CondeGil:
Add a text field to the UI settings to add an Alchemy api key to be used in the future to get the current account token list and balances. This is to be only used in live networks.

image


naps62:
Consider this approved, but will keep on hold until the rest of the alchemy stuff (backend) is merged 👍🏼

naps62 (approved):

PR #219, opened 2023-06-01
Live networks balances auto request

CondeGil:
- Move balances to a react context;
- Request tokens on wallet address change;
- Request tokens on network change;
- Request tokens balances every 10 min;
- Refactor wagmiWrapper to avoid side-effect config, and to be chain-aware. The wagmi wooks now need to receive the chainId, but the wrapper wont rebuild it self on network change;
- Split Networks context into two: Networks and currentNetwork. This is not that important but, splitting those two, makes things a bit more clear. On the frontend they do not really depend from each other, can avoid unnecessary renders and, because we will change the "current network" behavior in the future can be helpful on that effort.

### DIFF
--- a/gui/src/components/Balances.tsx
+++ b/gui/src/components/Balances.tsx
@@ -1,5 +1,6 @@
 import { Stack, Typography } from "@mui/material";
 import { erc20ABI } from "@wagmi/core";
+import { useEffect, useState } from "react";
 import { formatUnits } from "viem";
 import { useBalance, useContractRead } from "wagmi";
 
@@ -10,29 +11,19 @@ import { CopyToClipboard } from "./CopyToClipboard";
 import Panel from "./Panel";
 
 export function Balances() {
-  const address = useAccount();
-  const { balances } = useTokensBalances();
-
   return (
     <Panel>
       <Stack>
-        {address && <BalanceETH address={address} />}
-        {balances.map(([contract, balance]) => (
-          <BalanceERC20
-            key={contract}
-            {...{
-              contract,
-              balance: BigInt(balance),
-            }}
-          />
-        ))}
+        <BalanceETH />
+        <BalancesERC20 />
       </Stack>
     </Panel>
   );
 }
 
-function BalanceETH({ address }: { address: Address }) {
-  const { data: balance } = useBalance({ address });
+function BalanceETH() {
+  const address = useAccount();
+  const { data: balance } = useBalance({ address, enabled: !!address });
 
   if (!balance) return null;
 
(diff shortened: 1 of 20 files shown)


gui/src/components/ProviderNetworks.tsx
```diff
@@ -62,10 +50,12 @@ export function ProviderNetworks({ children }: { children: ReactNode }) {
...
         name: network.name,
         parent: actionId,
-        perform: () => value.setCurrentNetwork(network.name),
+        perform: () => {
+          setCurrentNetwork(network.name);
+        },
       })),
     ],
-    [networks, value.setCurrentNetwork]
+    [networks, setCurrentNetwork]
```
naps62, line 58: why is `setCurrentNetwork` still here if it's on `ProviderCurrentNetwork` as well?

gui/src/components/ProviderTokensBalances.tsx
```diff
@@ -0,0 +1,63 @@
...
+  >("db_get_erc20_balances", { chainId, address });
+
+  const { mutate: refetchTokensBalances } = useInvoke(
+    "alchemy_fetch_balances",
+    { chainId, address },
+    {
+      refreshInterval: 10 * 1000 * 60,
+      revalidateOnFocus: false,
+      revalidateOnMount: false,
+    }
```
naps62, line 35: nice, didn't even consider that SWR could support this already, but looks like a fine way to do it 👍🏼

tauri/src/alchemy/mod.rs
```diff
@@ -77,6 +80,7 @@ impl Alchemy {
             let res = (res["result"]).clone();
             let res: AlchemyResponse = serde_json::from_value(res)?;
             self.db.save_balances(res, chain_id).await?;
+            self.window_snd.send(Notify::TxsUpdated.into())?;
```
naps62, line 83: Actually for this we should add a new `BalancesUpdated` instead. we'll also eventually monitor transactions via alchemy, but that's a separate request

gui/src/components/ProviderNetworks.tsx
CondeGil: you need that for the kbar. But yeah, here is just a bug. the current network is yet to be instantiated at this point. Ill fix it

gui/src/components/Balances.tsx
```diff
@@ -49,23 +38,44 @@ function BalanceETH({ address }: { address: Address }) {
...
     functionName: "symbol",
+    chainId,
   });
 
   const { data: decimals } = useContractRead({
     address: contract,
     abi: erc20ABI,
     functionName: "decimals",
+    chainId,
   });
```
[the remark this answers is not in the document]
CondeGil: Just to make it clear, Are you talking of fetching the decimals and symbol on the tauri side?

tauri/src/alchemy/mod.rs
CondeGil: in the block_listener we're calling that as well.
image
What's the best way here? let the block_listener keep on only calling the TxsUpdated and listen both events on the frontend balancesProvider or make the balances provider only listening to BalancesUpdated and add two `self.window_snd.send` (TxsUpdated and BalancesUpdated) on the block_listener?

naps62: I think the second (send both events)

gui/src/components/Balances.tsx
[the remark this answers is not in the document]
naps62: yes

naps62 (approved):

PR #232, opened 2023-06-08
Live networks moved to backend

CondeGil:
- Move requests of decimals and symbol to backend
- Add new table to native token
- fetch native token info when needed

### DIFF
--- a/migrations/07_add_token_info_to_balances.sql
+++ b/migrations/07_add_token_info_to_balances.sql
@@ -1,0 +1,5 @@
+ALTER TABLE balances
+ADD symbol VARCHAR;
+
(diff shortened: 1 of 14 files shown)


migrations/07_add_token_info_to_balances.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE balances
+ADD symbol VARCHAR;
+
+ALTER TABLE balances
+ADD decimals INTEGER NOT NULL DEFAULT 0;
```
naps62, line 5: I think this should be a separate table. otherwise we'll have to duplicate this information all over

PR #237, opened 2023-06-12
Feature/native balances

CondeGil:
Decided to split this work into two branches. This one only contains the work of moving the native balance from the frontend to the backend. A future one will have the fetching erc20 metadata from the frontend to the backend.
- Add new table to native token
- fetch native token info when needed


gui/src/components/BalancesList.tsx
```diff
@@ -32,17 +32,16 @@ export function BalancesList() {
...
-
-  if (!balance) return null;
+  const { balance } = useNativeBalance();
 
-  return <BalanceItem balance={balance.value} decimals={18} symbol="ETH" />;
+  return (
+    <BalanceItem
+      balance={balance}
+      decimals={currentNetwork?.decimals || 18}
+      symbol={currentNetwork?.currency || "ETH"}
```
naps62, line 42: I think instead of this we can just do `if (!currentNetwork) return null;` right?

tauri/src/alchemy/mod.rs
```diff
@@ -94,4 +97,15 @@ impl Alchemy {
 
         Ok(serde_json::from_value(res["result"].clone())?)
     }
+
+    async fn fetch_native_balance(&self, chain_id: u32, address: Address) -> Result<()> {
+        let provider = Networks::write().await.get_current_provider();
+        let client = Arc::new(provider);
```
naps62, line 103: do you need the `Arc` here?

gui/src/App.tsx
```diff
@@ -28,30 +29,32 @@ export default function App() {
...
+                            <Route>
+                              <CommandBar>
+                                <Navbar />
+                                <HomePage />
+                              </CommandBar>
+                            </Route>
+                          </Switch>
+                        </Router>
+                      </ProviderTokensBalances>
+                    </ProviderNativeBalance>
```
naps62, line 57: ok, fine, but this is the last context we add before moving to a proper store. this is getting unwieldy 😅

PR #269, opened 2023-06-27
Remove ethers-core dependency

ZePedroResende:
This removes the redundant dependency on ethers-core, as we already have ethers.


naps62:
@ZePedroResende why so many changes on cargo.lock? sounds like you're including here some updates that may have happened when you pointed ethers to the github revision instead?

PR #308, opened 2023-07-22
fix: go submenu adding unexistent item

jopedroliveira:
**Why:**
Screenshot 2023-07-23 at 00 14 05

**How:**
Following up on the changes in https://github.com/iron-wallet/iron/commit/3541f1f9362abc05182c52df14e6fed069780bcc this PR removes the `details` item which was being added to the _go sub menu_ but not instantiated.


naps62:
damn feature flags tripping me up :D

PR #313, opened 2023-07-27
NFT view

### LINKED ISSUE #226: NFT view
We're already indexing ERC721s (on anvil at least). we need to create a view to display them for the user

jopedroliveira:
Closes #226 

This implements a new tab for listing all the wallet's ERC721 contracts.

- [x] FE Tab and list
- [x] FE Store
- [ ] BE DB function to fetch ERC721


jopedroliveira:
@naps62 This was as far as I could go. I may forgot something on the backend thus it gets nothing on the frontend. I'll keep looking at it, in the meantime if you have any advice plese shoot

.tool-versions
```diff
@@ -1 +1 @@
-nodejs 18.15.0
```
naps62, line 1: was there a reason for this? I was trying to keep it LTS

Cargo.lock
```diff
@@ -2808,6 +2808,7 @@ version = "0.4.1"
 dependencies = [
  "ethers",
  "iron-types",
+ "log",
```
naps62, line 2811: you may need to merge `main` back here. this isnt' supposed to be here I think

bin/iron/src/app.rs
```diff
@@ -41,6 +41,7 @@ impl IronApp {
                 iron_db::commands::db_get_contracts,
                 iron_db::commands::db_get_erc20_balances,
                 iron_db::commands::db_get_native_balance,
+                iron_db::commands::db_get_nft_tokens,
```
naps62, line 44: can we use `erc721` instead of `nft` across the board? nft is a bit ambiguous

crates/db/Cargo.toml
```diff
@@ -18,3 +18,5 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 tokio = { workspace = true }
 thiserror = { workspace = true }
+log = { version = "0.4", features = ["max_level_debug", "release_max_level_warn"] }
```
naps62, line 21: hmm, why? 🤔 

I have the `tracing` crate on `main` btw, which is a replacement for this

crates/db/src/commands.rs
```diff
@@ -1,5 +1,5 @@
 use ethers::types::{Address, U256};
-use iron_types::{events::Tx, TokenBalance};
+use iron_types::{events::Tx, TokenBalance, NftToken};
```
naps62, line 2: same here. `Erc721`. I probably need to rename some stuff regarding the erc20 ones as well, but that's on me

crates/db/src/lib.rs
```diff
@@ -292,6 +294,24 @@ impl DB {
...
+        r#" SELECT * 
+        FROM nft_tokens
+        WHERE chain_id = ?"#,
+      )
+      .bind(chain_id)
+      .map(|row| NftToken {
+        contract: Address::from_str(row.get::<&str, _>("contract")).unwrap(),
+        token_id: U256::from_dec_str(row.get::<&str, _>("token_id")).unwrap(),
+        owner: Address::from_str(row.get::<&str, _>("owner")).unwrap(),
+      })
```
naps62, line 308: instead of converting here, this should be done by implementing `impl TryFrom<SqliteRow> for NftToken`. check `tokens.rs` for how it was done for the existing ones

crates/types/src/tokens.rs
```diff
@@ -43,3 +43,13 @@ impl TryFrom<SqliteRow> for TokenMetadata {
         })
     }
 }
+
+#[derive(Debug, Serialize)]
+pub struct NftToken {
+    // pub contract: ChecksummedAddress,
```
naps62, line 49: can the comments be deleted?

gui/src/components/Nfts.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const loadMore = () => {
+    let pagination: Pagination = {};
+    const last = pages?.at(-1)?.pagination;
+    if (!!last) {
+      pagination = last;
+      pagination.page = (pagination.page || 0) + 1;
+    }
+
+    invoke<NftToken>("db_get_nft_tokens", {
```
naps62, line 36: instead of a direct db query here, we should create a `store/nfts.ts` to manage all the state

gui/src/components/Nfts.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  function NftGalleryItem({
+    contract,
+    imgSrc,
+    clickHandler,
+  }: {
+    contract: Address;
+    imgSrc: string | undefined;
+    clickHandler: () => void;
+  }) {
```
naps62, line 108: can we have a `interface NftGalleryItemProps`? I find it really hard to parse function headers this way

naps62 (commented):
Left a few comments. very good progress so far :)

crates/db/Cargo.toml
jopedroliveira: committed by mistake

Cargo.lock
jopedroliveira: committed by mistake

.tool-versions
jopedroliveira: committed by mistake

jopedroliveira:
@naps62 I'm facing trouble getting data from the backend and can't figure out why. 
I think there are two options:
- the nft_token table is empty and there's an issue somewhere on the transfers' parsing
- the mechanism that I implemented here is wrong
 
 Would you mind to take a look?

jopedroliveira:
@naps62 I'll close this PR because I'm opening two different ones, one for backend and another one for frontend

PR #314, opened 2023-07-27
Implement basic price feeds

### LINKED ISSUE #309: construct a list of price oracles
We want to compile a list of known price feeds, possibly using [chainlink's own listings](https://data.chain.link/ethereum/mainnet)
This list should be built at build time, or manually via a script.

This list can then be used to query exchange rates periodically (will be documented on a separate issue)

jmoreira-ls:
Fixes #309.

It embeds the JSON of price feeds to the binary in order for the executable to work on its own. It also has a custom `Feed` deserializer in order to be able to add multiple types of price feeds to the JSON.


crates/prices/src/init.rs
```diff
@@ -0,0 +1,28 @@
...
+
+static FEEDS: OnceCell<RwLock<Feeds>> = OnceCell::new();
+
+pub async fn init() {
+    // Embed the JSON file in the binary
+    let feeds_str = include_str!("../res/feeds.json");
+
+    let res: Feeds = serde_json::from_str(feeds_str).unwrap();
+
+    FEEDS.set(RwLock::new(res)).unwrap();
```
naps62, line 16: instead of `unwrap`s, this should just go with the `?` operator, and return a `crate::Result<()>`

crates/prices/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "iron-prices"
```
naps62, line 2: maybe `iron-exchange-rates` would be a better name now that I think about it. what do you think?

crates/prices/src/feed.rs
```diff
@@ -0,0 +1,55 @@
...
+        A: serde::de::SeqAccess<'de>,
+    {
+        let provider = seq.next_element()?.unwrap();
+        let address = seq.next_element()?.unwrap();
+        match Feed::create(provider, address) {
+            Ok(value) => Ok(value),
+            Err(other) => return Err(de::Error::custom(other)),
+        }
+    }
+}
```
naps62, line 29: I'm pretty sure you didn't have to implement your own deserializer 😅 usually you should be able to go with just `#[derive(Deserialize)]`. if you can't, it either means you're doing some more complex parsing (not the case here), or you're not setting the format properly

crates/prices/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+pub use init::init;
+
+use feed::Feed;
+use serde::Deserialize;
+use std::collections::HashMap;
+
+#[derive(Deserialize, Debug)]
+pub struct Feeds {
+    // Structure for feed discovery: network -> currency -> asset
+    feeds: HashMap<String, HashMap<String, HashMap<String, Vec<Feed>>>>,
```
naps62, line 15: the last one should be a `ethers::types::Address`, not a String

crates/prices/res/feeds.json
```diff
@@ -0,0 +1,54 @@
+{
+    "feeds": {
+        "mainnet": {
+            "USD": {
+                "1INCH": [
+                    ["chainlink", "0xc929ad75B72593967DE83E7F7Cda0493458261D9"]
+                ],
```
naps62, line 7: I think something like this would make more sense:

```
1INCH: {
  chainlink: "0x....."
},
AAVE: {
  chainlink: "0x..."
}
```

this is probably why you weren't able to just derive the Deserializer

crates/prices/Cargo.toml
jmoreira-ls: sounds good!

crates/prices/src/feed.rs
jmoreira-ls: my objective was to automatically deserialize into one of the `Feed` enum depending on the first string of the vector.  I'll take a better look on how to do this then.

crates/prices/res/feeds.json
jmoreira-ls: that looks better, like I said previously I'm going to dive a deeper on the deserialization to get a better solution.

crates/prices/src/lib.rs
jmoreira-ls: definitely looks like the best approach, will change it.

jmoreira-ls: regarding the `ethers::types::address`, do you mean asset instead? for currency I was considering "USD" and asset I was assuming "ETH", so I guess it's the asset that should be `Address`?

crates/prices/res/feeds.json
jmoreira-ls: I ended up using internal tags for derialization, let me know if you agree.

crates/prices/src/lib.rs
jmoreira-ls: I ended up keeping the `String` as it made sense to me using the token symbol instead, as it is in the JSON, to keep the feeds file more readable (token symbol vs address).

Another option would be, while deserializing, to translate the token symbol to its address depending on the network, but I think it's simpler just to access the token metadata and retrieve the symbol before accessing the feed.

Let me know if you disagree!

crates/exchange-rates/res/feeds.json
```diff
@@ -0,0 +1,96 @@
...
+                ],
+                "ETH": [
+                    {
+                        "provider": "chainlink",
+                        "address": "0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419"
+                    }
+                ]
+            }
+        },
+        "goerli": {
```
naps62, line 79: and similarly here. goerli is `5`

crates/exchange-rates/src/init.rs
```diff
@@ -0,0 +1,47 @@
...
+        feeds: HashMap::new()
+    };
+
+    for (k1, v1) in res.feeds {
+        for (k2, v2) in v1 {
+            for (k3, v3) in v2 {
+                feeds.feeds.insert((k1.clone(), k2.clone(), k3.clone()), v3);
+            }
+        }
+    }
```
naps62, line 33: this whole thing should be unnecessary. this should just work:

```rust
let feeds: Feeds = serde_json::from_str(feeds_str).unwrap();
```

if it doesn't, it means the json and the serde derives are not matching correctly

crates/exchange-rates/src/init.rs
```diff
@@ -0,0 +1,47 @@
...
+use iron_types::GlobalState;
+use once_cell::sync::OnceCell;
+use serde::Deserialize;
+use std::collections::HashMap;
+use tokio::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
+
+use super::feed::Feed;
+use super::Feeds;
+
+static FEEDS: OnceCell<RwLock<Feeds>> = OnceCell::new();
```
naps62, line 11: instead of `OnceCell` and an `init` function, this can be simplified with `Lazy`

static FEEDS: Lazy<RwLock<Feeds>> = Lazy::new(|| {
  let feeds: Feeds = ...;

  RwLock::new(feeds)
});
```

the custom init function is only needed when we need additional arguments

File: crates/exchange-rates/src/feed.rs [hunk 9]
```diff
@@ -0,0 +1,8 @@
+use ethers::types::Address;
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+#[serde(tag = "provider", rename_all = "camelCase", content = "address")]
```
naps62, line 5: hmm, didn't even know about the `content` attr, but seems exactly what we want. nice one

PR #315, opened 2023-07-28
Create onboarding wizard

paulacarneiro:
Create onboarding wizard to help first time users navigate the app.
- uses custom store to keep track of user input
- stores onboarding state in `localStorage`
 
Current steps:
- Introduction with link to github repo
- Live blockchain, where users can add their Alchemy API key
- Final considerations

Missing steps:
- Anvil default mnemonic opt-out option
- Check browser extension


gui/src/components/OnboardingWizard/OnboardingWrapper.tsx
```diff
@@ -0,0 +1,22 @@
+import { useState } from "react";
+
+import { OnboardingWizard } from "./OnboardingWizard";
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export function OnboardingWrapper({ children }: Props) {
+  const onboarded = localStorage.getItem("onboarded");
```
gabrielpoca, line 10: This may not be a good idea. As far as I remember, reading from the localStorage is a sync operation; in this case, you do it in every render. You can push this outside of react because you only need to read from the localstorage when you start the app.

gui/src/components/OnboardingWizard/OnboardingWizard.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useWizardForm } from "../../store/wizard";
+import { Logo } from "../Logo";
+import { OnboardingCarousel } from "./OnboardingCarousel";
+import { steps } from "./Steps";
+
+interface Props {
+  closeOnboarding: () => void;
+}
+
+export function OnboardingWizard({ closeOnboarding }: Props) {
```
gabrielpoca, line 15: I noticed you added everything inside a folder `OnboardingWizard`, but I think it would be best to push everything to the `components` folder and use the prefix `Onboarding` to maintain consistency with the rest of the application.

gui/src/components/OnboardingWizard/OnboardingCarousel.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { useTheme } from "@mui/material/styles";
+
+import { Step } from "./Steps";
+
+interface Props {
+  steps: Step[];
+  activeStep: number;
+  handleClose: () => void;
+  handleNext: () => void;
+  handleBack: () => void;
```
gabrielpoca, line 16: I believe `unknown` is the recommended approach, instead of `void`. Right?

gui/src/components/OnboardingWizard/OnboardingWizard.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const [activeStep, setActiveStep] = useState(0);
+  const { alchemyApiKey } = useWizardForm();
```
gabrielpoca, line 17: This feels like unnecessary complexity. What about reading directly from the settings with `  const { data: general, mutate } = useInvoke<GeneralSettings>("settings_get");`?

gabrielpoca (changes requested):
Looks good :+1:

gui/src/components/OnboardingWizard/OnboardingWizard.tsx
paulacarneiro: That makes sense to read the initial value. I created the store to save the user inputs in case the user moves back and forth between the steps, and only apply these changes when the wizard is closed

gui/src/components/OnboardingWizard/OnboardingWizard.tsx
naps62: Actually I was in the process of breaking that consistency for all the Settings* stuff, because it was making my fuzzy finding a pain to use

gui/src/components/OnboardingWizard/OnboardingWizard.tsx
paulacarneiro: I added this data read and got rid of the store, the form data is now being stored in a state variable in `OnboardingWizard`

gui/src/components/OnboardingWizard/OnboardingCarousel.tsx
gabrielpoca: But that's the point; you don't care if this function returns something, so why force it to return void? The beauty of `unknown` is that even if the function returns something, you can't do anything with the return value.
For instance, the parent component may also use the `handleNext` for other components where it has to return something. By setting the return type to void, you don't allow for that, even if it doesn't matter for your component.

gui/src/components/OnboardingWizard/OnboardingWizard.tsx
gabrielpoca: 👍 it's fine by me, but in that case, be very careful where you place components, and maybe don't allow for another level of nesting. It's really easy for nesting to become a mess.

gui/src/components/OnboardingWizard/OnboardingWrapper.tsx
gabrielpoca: @naps62 do you think it matters? The localstorage is also persistent

gui/src/components/OnboardingWizard/OnboardingCarousel.tsx
naps62: semantically, `void` makes more sense here. `unknown` doesn't mean you can't do anything with it. you can still assign the result to a binding or put it in an object thinking that it's not null/undefined

in practice the only problem is it forces you to not return anything, meaning that `() => foo()` has to actually be `() => { foo(); }`, which is more cumbersome

gui/src/components/OnboardingWizard/OnboardingWrapper.tsx
naps62: Having it in your `settings.json` means you can safely port your settings to a new computer and not have the onboarding jump on you when you already have done it

gui/src/components/OnboardingWizard/OnboardingCarousel.tsx
gabrielpoca: What do you mean? I don't understand what you said. [You cannot use the return value of a function that returns 'unknown'](https://www.typescriptlang.org/docs/handbook/2/functions.html#unknown). That's the goal, really; you don't care what the function returns, no matter if it returns something or nothing at all, you don't care. When you use void, you are explicitly saying that you care. In my opinion, in this scenario, you shouldn't care.

gui/src/components/OnboardingWizard/OnboardingWrapper.tsx
gabrielpoca: 👍, can you help @paulacarneiro make the change?

gui/src/components/OnboardingWizard/OnboardingCarousel.tsx
naps62: That's not true. we're getting really into semantics here, but: unknown is exactly what the name suggests. it's unknown. you can't call method functions on it because you don't know which ones exist. but you can still use the value:

```
function u(): unknown {
    return 1
}
// this works
const x = u();
console.log({ foo: x});
```

the semantics of `void` are precisely to prevent you from even having `return 1` in the first place
saying "I explicitly don't return anything" is very different from saying "I may return something, but refuse to tell you what it is"

gui/src/components/OnboardingWizard/OnboardingWrapper.tsx
naps62: yes. will take care of it tomorrow

naps62: @paulacarneiro here you go: https://github.com/iron-wallet/iron/pull/322/files

feel free to merge this locally to your branch, or just let me know and I'll merge directly from the PR

with that code, you should be able to use `invoke("settings_get")` (which already existed, you can find it's usage in the code, but it now has a new `onboarded` boolean), and `invoke("settings_finish_onboarding")` to set it to true

gui/src/components/OnboardingWizard/OnboardingCarousel.tsx
gabrielpoca: Interesting, I'm not sure why I thought you couldn't use an unknown, but it makes sense:

```ts
function one() {
    return "world"
}

function two(fn: () => unknown) {
    const value = fn();

    if (typeof value === 'string') {
        console.log("It's a string!")
    }
}

two(one)
```

Do you know what the funny part is? You can do the same with a void:

```
function one() {
    return "world"
}

function two(fn: () => void) {
    const value = fn();

    if (typeof value === 'string') {
        console.log("It's a string!")
    }
}

two(one)
```

So now I no longer know what to think 😅

naps62: hmm, didn't know that weird trick worked. my experience with `void` is only that TS prevents you from having a `return` statement inside the function (or an implicit return with arrow functions)

but apparently you can ask eslint to catch that, at least:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-confusing-void-expression.md

gabrielpoca: Nice, we should enable that. For the sake of this change, if having here void doesn't require the `() => { foo(); }` on the parent, I think it's ok, if not, I still think it's probably better to use the unknown.

gabrielpoca (approved):
Looks good to me! @naps62 ?

naps62:
will try now

gui/src/components/Onboarding/Wrapper.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { useInvoke } from "../../hooks";
+import { GeneralSettings } from "../../types";
+import { OnboardingWizard } from "./Wizard";
+
+interface Props {
+  children: React.ReactNode;
+}
+
+export function OnboardingWrapper({ children }: Props) {
```
naps62, line 12: will leave this for future work but: I'd rather see the onboarding implemented as a separate window (and configuring the app to choose on boot which one to load) than having this arbitrary wrapping of the entire app

it also means we can have some better control over e.g. window sizes: we may change the main window's size in the future, and that may not translate well to this onboarding

naps62 (approved):

naps62:
nice work @paulacarneiro just changed a couple things:
- rewrote some of the copy
- moved everything to a `components/Onboarding/`. I know gabriel doesn't like that, but we don't need to tell him :D

gui/src/components/OnboardingWizard/OnboardingCarousel.tsx
Hari-Bombon: @naps62  I will take this one . Gimme the access

Hari-Bombon: @naps62  I will take this one . Gimme the access

Hari-Bombon:
@naps62  I will take this one naps. Gimme the access

PR #321, opened 2023-08-03
Add Sidebar

gabrielpoca:
In this change, we replace the tabs with a sidebar. The goal is to make Iron look more like any other desktop app while at the same time making the UI more self-explanatory.

We are also moving the dropdown selectors for wallet, address, and network to the sidebar. They have labels to make them easy to figure out, and the style change so they are less "present" in the UI.

The sidebar also has a compact mode, where the text and dropdown disappear, and we are left with just icons. This mode is nice for those who don't want Iron taking up a lot of space and only need to look up small bits of information or perform short interactions. The dropdowns disappear, but we still have the button for the command bar that can be used to click around and perform actions like changing network, wallet, or address.

The navbar also got some updates to make it look more like a native app.

There are other minor fixes across the codebase.

Screenshot 2023-08-03 at 18 00 39
Screenshot 2023-08-03 at 18 01 02

### DIFF
--- a/gui/src/components/Sidebar.tsx
+++ b/gui/src/components/Sidebar.tsx
@@ -1,0 +1,172 @@
+import {
+  Box,
+  Drawer,
+  List,
+  ListItemButton,
+  ListItemIcon,
+  ListItemText,
+} from "@mui/material";
+import { findIndex } from "lodash-es";
+import { Link, useLocation, useRoute } from "wouter";
+import ArrowOutwardSharpIcon from "@mui/icons-material/ArrowOutwardSharp";
+import RequestQuoteSharpIcon from "@mui/icons-material/RequestQuoteSharp";
+import TerminalSharpIcon from "@mui/icons-material/TerminalSharp";
+import OnlinePredictionSharpIcon from "@mui/icons-material/OnlinePredictionSharp";
+import { parseInt, range, toString } from "lodash-es";
+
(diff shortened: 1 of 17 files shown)


gui/src/components/Sidebar.tsx
```diff
@@ -0,0 +1,216 @@
...
+  );
+}
+
+function SidebarTab({
+  tab,
+  selected,
+}: {
+  tab: (typeof TABS)[number];
+  selected: boolean;
+}) {
```
naps62, line 168: can you move this to a `SidebarTabProps` interface?

naps62:
just the one comment, but otherwise lgtm

gui/src/components/Sidebar.tsx
gabrielpoca: I can, but do you mind me asking why? I'm not against it; just trying to understand your reasoning. I usually only export these to interfaces when they become messy.

naps62: I don't mean exporting it, just declaring it separately from the component.
I find this syntax honestly terrible to read. it's fine'ish when it's just one element (such as `{ children: React.Node }`), but anything more becomes really ugly

gabrielpoca: 👍

naps62 (approved):
:shipit:

PR #325, opened 2023-08-10
Erc721 list - backend

### LINKED ISSUE #226: NFT view

jopedroliveira:
Partially closes #226 

This PR implements the backend logic for storing ERC721 tokens and collections. 
This includes two new tables on the database, one for each token of an ERC721 and one for the ERC721 contract itself (collections). 

Also implements two APIs, one for fetching tokens
```json
{   
 "contract": "0x5fbdb2315678afecb367f032d93f642f64180aa3",
 "metadata": "{id:1,name:1,image:https://assets.pokemon.com/assets/cms2/img/pokedex/detail/001.png}",
 "owner": "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
 "token_id": "0x1",
 "uri": "data:application/json;base64,eyJpZCI6MSwibmFtZSI6IjEiLCJpbWFnZSI6Imh0dHBzOi8vYXNzZXRzLnBva2Vtb24uY29tL2Fzc2V0cy9jbXMyL2ltZy9wb2tlZGV4L2RldGF…"
"metadata": {"id":1,"name":"1","image":"https://assets.pokemon.com/assets/cms2/img/pokedex/detail/001.png"},
 "name": "POKEMON",
 "symbol": "PKMN"
}
 ```

### CODE REVIEW

File: migrations/11_create_nfts_metadata.sql [hunk 1]
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE nfts_metadata (
```
naps62, line 1: can we rename to erc721_metadata?

jopedroliveira: Yes. Can we rename the existing table of `nft_tokens` as well?

migrations/11_create_nfts_metadata.sql
```diff
@@ -0,0 +1,9 @@
...
+  contract VARCHAR COLLATE NOCASE,
+  chain_id INTEGER NOT NULL,
+  token_id VARCHAR NOT NULL,
+  name VARCHAR NOT NULL,
+  symbol VARCHAR NOT NULL,
+  uri VARCHAR NOT NULL,
```
naps62, line 7: how about fetching & including the actual json metadata that comes from this URL?

crates/types/src/tokens.rs
```diff
@@ -43,3 +43,50 @@ impl TryFrom<SqliteRow> for TokenMetadata {
...
+
+impl TryFrom<SqliteRow> for Erc721Token {
+    type Error = ();
+
+    fn try_from(row: SqliteRow) -> Result<Self, Self::Error> {
+        Ok(Self {
+            name: row.get("name"),
+            symbol: row.get("symbol"),
+            uri: row.get("uri"),
+            info: row.try_into().unwrap(),
```
naps62, line 89:
```suggestion
            info: row.try_into()?,
```

we should avoid unwraps, especially in these contexts

crates/sync/anvil/src/tracker.rs
```diff
@@ -215,6 +215,20 @@ async fn process(ctx: Ctx, mut block_rcv: mpsc::UnboundedReceiver<Msg>) -> Resul
...
+            .await?
+            .into_iter()
+        {
+            let token_id = erc721_token_info.token_id;
+            let address = erc721_token_info.contract;
+            let metadata = fetch_erc721_metadata(erc721_token_info, &provider).await;
+            ctx.db
+                .save_erc721_metadata(address, ctx.chain_id, token_id, metadata)
+                .await?;
+        }
```
naps62, line 230: now that I re-read this and the db table, this could be a problem.
I expected the `erc721` database table to have 1 record per collection. but you're storing 1 record per item, which will scale up these requests considerably, even though they're mostly duplicate (only the token URI differs from each one)

so I'd propose something different, which is to have 2 tables:
one for `erc721_collections` (which tracks name & symbol) and one for `erc721_metadata` (which tracks individual token_ids and their URI)

naps62: to be clear. the problem is not so much with this anvil side, but with how the current data schema will translate to alchemy.com. we can't afford making that many redundant requests for every individual item

migrations/11_create_nfts_metadata.sql
naps62: ohhh, we already had one? 🤔 
then a lot of comments here need to reviewed

naps62:
@jopedroliveira sorry for the confusion, I guess I forgot we already had a `nft_tokens` table.

> The pre-existing db was already storing ERC721 data by token_id, so, could that table also store the metadata, namely name, symbol, and URI? To make it consistent with ERC20 I didn't merge it

let's clarify naming here: name symbol and URI are not the same kind of metadata. Usually when we talk about metadata in the context of an NFT, we're talking about the individual JSON metadata that each item has (stored off-chain). name and symbol are metadata of the contract itself, and are the same for the entire collection.

so my overall proposal is to have:

table `erc721_collections`:
- contract_address
- chain_id
- name
- symbol

table `erc721_tokens`:
- contract_address
- chain_id
- token_id
- owner
- uri
- metadata (optional, we can also deal with this one in a future PR to make it simpler)

I guess it's easier if you just drop the existing `nft_tokens` table instead of trying to figure out how to fit it in

crates/db/src/lib.rs
```diff
@@ -308,6 +307,82 @@ impl DB {
...
+    pub async fn save_erc721_data(
+        &self,
+        address: Address,
+        chain_id: u32,
+        token_id: U256,
+        owner: Address,
+        name: String,
+        symbol: String,
+        uri: String,
+        metadata: String,
```
jopedroliveira, line 371: Linter is showing an error here
`error: this function has too many arguments (9/7)`

crates/sync/anvil/src/tracker.rs
```diff
@@ -249,3 +277,42 @@ pub async fn fetch_erc20_metadata(address: Address, client: &Provider<Http>) ->
...
+        md = String::from_utf8(decoded_uri.unwrap()).unwrap();
+    } else if contract_uri.contains("ipfs://") {
+        let contract_uri = contract_uri.replace("ipfs://", "https://ipfs.io/ipfs/");
+        let response = reqwest::get(contract_uri.clone()).await.unwrap();
+
+        if response.status().is_success() {
+            md = response.text().await.unwrap();
+        }
+    } else {
+        // TODO: handle exceptions
```
jopedroliveira, line 309: If the tokenURI is nor base64 encoded, nor an ipfs URL, then it should be handled differently.

crates/db/src/commands.rs
```diff
@@ -44,8 +44,17 @@ pub async fn db_get_native_balance(
...
-    address: Address,
+    owner: Address,
     db: tauri::State<'_, DB>,
 ) -> Result<Vec<Erc721Token>> {
-    db.get_erc721_tokens(chain_id, address).await
+    db.get_erc721_tokens(chain_id, owner).await
+}
+
+#[tauri::command]
+pub async fn db_get_erc721_collections(
```
naps62, line 54: what is the need for this command?

crates/db/src/queries.rs
```diff
@@ -115,44 +115,64 @@ pub(super) fn erc20_update_balance<'a>(
...
-        .bind(chain_id)
         .bind(format!("0x{:x}", tx.contract))
+        .bind(chain_id)
         .bind(format!("0x{:x}", tx.token_id))
         .bind(format!("0x{:x}", tx.to))
     }
 }
 
-pub(super) fn update_erc721_metadata<'a>(
+// TODO - refactor
```
naps62: what's there to refactor?

crates/db/src/queries.rs
```diff
@@ -115,44 +115,64 @@ pub(super) fn erc20_update_balance<'a>(
...
         .bind(chain_id)
         .bind(format!("0x{:x}", tx.contract))
         .bind(format!("0x{:x}", tx.token_id))
     } else {
         // minting or transfer
         sqlx::query(
-            r#" INSERT OR REPLACE INTO nft_tokens (chain_id, contract, token_id, owner)
-                        VALUES (?,?,?,?)"#,
+            r#" INSERT OR REPLACE INTO erc721_tokens (contract, chain_id, token_id, owner, uri, metadata)
+                        VALUES (?,?,?,?, null, null)"#,
```
naps62, line 131: if you're inserting null, why not just remove the `uri, metadata` part from the list of fields?

crates/sync/anvil/src/tracker.rs
```diff
@@ -265,18 +279,43 @@ pub async fn fetch_erc20_metadata(address: Address, client: &Provider<Http>) ->
...
 
-pub async fn fetch_erc721_metadata(
-    erc721_token_info: Erc721TokenInfo,
+pub async fn fetch_erc721_data(
+    erc721_token: Erc721Token,
     client: &Provider<Http>,
-) -> Erc721TokenMetadata {
-    let contract = IERC721::new(erc721_token_info.contract, Arc::new(client));
-    Erc721TokenMetadata {
+) -> Erc721TokenData {
```
naps62, line 284: This needs to return `Result<Erc721TokenData>`, and we can't do `unwrap()`'s anywhere inside, since that'll completely break the tracker

you can refer to david's PR to see how this should be haandled (question mark operator), or you can ping me next week

crates/sync/anvil/src/tracker.rs
```diff
@@ -1,5 +1,6 @@
-use std::{sync::Arc, time::Duration};
+use std::{ops::Add, sync::Arc, time::Duration};
 
+use base64::{engine::general_purpose, Engine as _};
```
naps62, line 3: this should just be 

```
use base64::{self, Engine as _};
```

otherwise the call to `general_purpose` further down is hard to understand, since `general_purpose` by itself explains nothing

crates/sync/anvil/src/tracker.rs
```diff
@@ -215,17 +217,29 @@ async fn process(ctx: Ctx, mut block_rcv: mpsc::UnboundedReceiver<Msg>) -> Resul
...
         {
-            let token_id = erc721_token_info.token_id;
-            let address = erc721_token_info.contract;
-            let metadata = fetch_erc721_metadata(erc721_token_info, &provider).await;
+            let token_id = erc721_token.token_id;
+            let address = erc721_token.contract;
+            let owner = erc721_token.owner;
+            let erc721_data = fetch_erc721_data(erc721_token, &provider).await;
+
+            dbg!(&erc721_data);
```
naps62: ⚠️

crates/db/src/lib.rs
```diff
@@ -348,16 +359,23 @@ impl DB {
...
     ) -> Result<()> {
         let mut conn = self.tx().await?;
+        queries::update_erc721(address, chain_id, token_id, owner, uri, metadata)
+            .execute(&mut conn)
+            .await?;
 
-        queries::update_erc721_metadata(address, chain_id, token_id, metadata)
+        queries::update_erc721collection(address, chain_id, name, symbol)
```
naps62, line 378: these should be two separate calls: one to update a collection, and one to update a particular item. especially because they should happen with different frequencias (which is not how this currently does it)

crates/sync/anvil/src/tracker.rs
```diff
@@ -215,17 +217,29 @@ async fn process(ctx: Ctx, mut block_rcv: mpsc::UnboundedReceiver<Msg>) -> Resul
...
+                .save_erc721_data(
+                    address,
+                    ctx.chain_id,
+                    token_id,
+                    owner,
+                    erc721_data.name,
+                    erc721_data.symbol,
+                    erc721_data.uri,
+                    erc721_data.metadata,
+                )
```
naps62, line 241: This loop is once again mixing two concepts: the collection, and the individual item

`db.get_erc721_missing_data` only checks for missing collection data. meaning that it will skip items for which the collection exists, but the item metadata doesn't. that means they'll never get updated

These should happen as two independent tasks: update any collections. and updating any items

crates/sync/anvil/src/tracker.rs
jopedroliveira: Referring to [this](https://docs.rs/base64/latest/base64/#using-predefined-engines) here since this seems to be the new recommended way to decode `base64`.

crates/db/src/commands.rs
jopedroliveira: The idea was for the frontend to store collection data and avoid making multiple database reads for tokens from the same collection. This approach also helps in avoiding the need to create a specific type just to return the token and collection in a single request. Additionally, it could enable different displays on the frontend, such as having collections and then entering to view the tokens, or something similar

crates/sync/anvil/src/tracker.rs
naps62: I think you're reading too much into that snippet. it's by no means a recommendation on how to set the `use` statement.
The functionality is exactly the same, this is purely a matter of readability.

And in anything other than a small snippet, having a `general_purpose` keyword dangling in the code is just plain confusing, since one would have to jump up to the imports to understand where it came from

crates/db/src/lib.rs
```diff
@@ -327,6 +326,90 @@ impl DB {
         Ok(res)
     }
 
+    pub async fn get_erc721_tokens_missing_data(&self, chain_id: u32) -> Result<Vec<Erc721Token>> {
```
naps62, line 329:
```suggestion
    pub async fn get_erc721_tokens_with_missing_data(&self, chain_id: u32) -> Result<Vec<Erc721Token>> {
```

crates/db/src/lib.rs
```diff
@@ -327,6 +326,90 @@ impl DB {
...
+        let mut conn = self.tx().await?;
+        queries::update_erc721_token(address, chain_id, token_id, owner, uri, metadata)
+            .execute(&mut conn)
+            .await?;
+
+        conn.commit().await?;
+        Ok(())
+    }
+
+    pub async fn get_erc721_missing_collections(&self, chain_id: u32) -> Result<Vec<Address>> {
```
naps62, line 360:
```suggestion
    pub async fn get_erc721_with_missing_collections(&self, chain_id: u32) -> Result<Vec<Address>> {
```

naps62 (approved):
Since it's draft still I'm not sure if you want to merge?
but overall looks good, other than the couple comments I left

I haven't had a chance to test it for myself, but assuming you did. feel free to merge while I'm out ✌🏼

PR #326, opened 2023-08-10
Setup extension publishing on CI

naps62:
The secrets are already setup, but I'm waiting on approval from the chrome store before merging this


naps62:
Update: the chrome store plugin was accepted (after 3 attemps)
hopefully next submissions are faster (?) now I need to test an automated deploy via the CI

naps62:
update: waiting on firefox store approval

PR #329, opened 2023-08-11
Sidebars and Toolbars

gabrielpoca:
This change addresses multiple minor issues:

* Changed the settings to use a sidebar instead of tabs.
* Disabled `user-select` globally so the app behaves more like a native app.
* Removed the icon from the sidebar/toolbar on mac because it's not common to see them.
* Fixed the toolbar so it's also present in the sidebar to make sure the window can be dragged when someone clicks on the top of the sidebar.
* Addressed padding inconsistencies between macOS and other systems.
* Fixed the breaking points in the settings and command line buttons.

Screenshot 2023-08-11 at 12 25 54

### DIFF
--- a/gui/src/App.tsx
+++ b/gui/src/App.tsx
@@ -1,4 +1,4 @@
-import { ThemeProvider } from "@mui/material";
+import { GlobalStyles, ThemeProvider } from "@mui/material";
 import CssBaseline from "@mui/material/CssBaseline";
 import { QueryClient, QueryClientProvider } from "react-query";
 import { Route, Router, Switch } from "wouter";
@@ -18,11 +18,20 @@ const queryClient = new QueryClient({
   defaultOptions: { queries: { suspense: true } },
 });
 
+const globalStyles = {
+  body: { userSelect: "none" },
+  p: { userSelect: "initial" },
+  h1: { userSelect: "initial" },
+  h2: { userSelect: "initial" },
+  h3: { userSelect: "initial" },
+};
+
 export default function App() {
   const theme = useTheme((s) => s.theme);
 
   return (
     <ThemeProvider theme={theme}>
+      <GlobalStyles styles={globalStyles} />
       <CssBaseline>
         <QueryClientProvider client={queryClient}>
           <OnboardingWrapper>
--- a/gui/src/components/Onboarding/Wizard.tsx
+++ b/gui/src/components/Onboarding/Wizard.tsx
@@ -3,8 +3,7 @@ import { invoke } from "@tauri-apps/api/tauri";
 import { useEffect, useState } from "react";
 
 import { GeneralSettings } from "../../types";
-import { Logo } from "../Logo";
-import { useInvoke, useKeyPress, useOS } from "./../../hooks";
+import { useInvoke, useKeyPress } from "./../../hooks";
 import { OnboardingCarousel } from "./Carousel";
 import { steps } from "./Steps";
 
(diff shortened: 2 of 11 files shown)


gui/src/App.tsx
```diff
@@ -18,11 +18,20 @@ const queryClient = new QueryClient({
   defaultOptions: { queries: { suspense: true } },
 });
 
+const globalStyles = {
+  body: { userSelect: "none" },
+  p: { userSelect: "initial" },
+  h1: { userSelect: "initial" },
+  h2: { userSelect: "initial" },
+  h3: { userSelect: "initial" },
+};
```
naps62, line 27: Without testing, I'm assuming this means I can't select anything by default, but with exceptions added for `Typography`?

gui/src/hooks/useOS.ts
```diff
@@ -0,0 +1,14 @@
+import * as os from "@tauri-apps/api/os";
+import { useEffect, useState } from "react";
+
+export function useOS() {
+  const [type, setType] = useState<string | undefined>();
```
naps62, line 5: do we know the specific types this gets?
would be useful to use a more concrete type instead of a general string

gui/src/components/Onboarding/Wizard.tsx
```diff
@@ -51,10 +52,12 @@ export function OnboardingWizard({ closeOnboarding }: Props) {
...
-      <Stack direction="row" py={1.5} spacing={1} alignItems="center">
-        <Logo width={40} />
-        <Typography>Iron Wallet</Typography>
-      </Stack>
+      <Box data-tauri-drag-region="true" pt={4}></Box>
+      {type !== "Darwin" && (
+        <Box>
+          <Logo width={40} />
+        </Box>
+      )}
```
naps62, line 60: why not hide it globally? this feels like it's not a macOS specific thing

gui/src/components/SettingsButton.tsx
```diff
@@ -33,7 +33,7 @@ export function SettingsButton() {
         onClick={() => setShowSettings(true)}
         sx={{
           justifyContent: "flex-start",
-          [theme.breakpoints.down("md")]: {
+          [theme.breakpoints.down("sm")]: {
```
naps62, line 36: I'm assuming this is due to the change I made to the sidebar breakpoints.
would it make sense to add a custom `sidebar` breakpoint to make the dependency more clear?

gui/src/App.tsx
gabrielpoca: You can still interact with things, but things won't be "selectable". I'm not sure exactly how to explain it; it's like most apps do it. I can show it in a call

gui/src/components/Onboarding/Wizard.tsx
gabrielpoca: I don't know 🤷 in macos we already have the icon on the menubar, and we don't usually see it on the app itself. I'm not sure about how it works in other systems

gui/src/components/SettingsButton.tsx
gabrielpoca: I thought about that, but I'm also not sure. We only have a few breakpoints so I don't think it's a big issue

gui/src/components/Onboarding/Wizard.tsx
naps62: I was checking slack, telegram, obsidian and discord. The only one that shows the logo is Discord, but because it's the same button that goes to DMs

should be fine to remove I think

gui/src/hooks/useOS.ts
gabrielpoca: We can just use tauri's types 👍

gui/src/components/Onboarding/Wizard.tsx
gabrielpoca: 👍 done

naps62 (approved):

PR #332, opened 2023-08-11
Refactor contracts logic to backend

DavideSilva:


crates/contracts/src/commands.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+async fn fetch_abi(chain: Chain, address: Address) -> Result<Option<Metadata>> {
+    let api_key = std::env::var("ETHERSCAN_API_KEY")?;
+
+    let client = Client::new(chain, api_key)?;
+    let metadata = client.contract_source_code(address).await?;
+
+    Ok(Some(metadata.items[0].clone()))
+}
```
naps62, line 41: this doesn't feel worth its own crate, tbh
the commands could go directly to `iron-db`, and this `fetch_abi` function would probably be suitable for `iron-sync`

crates/contracts/src/commands.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    // self.window_snd.send(UINotify::BalancesUpdated.into())?;
+    // send ContractsUpdated event to UI using iron_broadcast
+
+    db.insert_contract(chain_id, address, abi, contract_name)
+        .await
+}
+
+async fn fetch_abi(chain: Chain, address: Address) -> Result<Option<Metadata>> {
+    let api_key = std::env::var("ETHERSCAN_API_KEY")?;
```
naps62, line 35: this can't come from an env variable. needs to be on `Settings` instead, and be configurable through the UI settings, just like it already happens for alchemy

PR #334, opened 2023-08-13
Better settings and sidebars

gabrielpoca:
In this change:

* Update the settings to use a drawer/sidebar and a title.
* Update the sidebar to make it look more like a native app.
* Updated the menu bar to include the standard "Settings.." menu that can be triggered using command+comma. I had to copy over the menu bar in the function `os_default`.
* Added a store for the settings and a global listener to toggle open/close the settings from the menu bar.

Screenshot 2023-08-13 at 10 53 33
Screenshot 2023-08-13 at 10 53 43
Screenshot 2023-08-13 at 10 54 01


gui/src/components/NewVersionNotice.tsx
```diff
@@ -49,16 +49,23 @@ export function NewVersionNotice() {
...
       open={open}
       onClose={handleClose}
       key={vertical + horizontal}
     >
-      <Link href="https://github.com/iron-wallet/iron/releases" target="_blank">
+      <Link
+        href="https://github.com/iron-wallet/iron/releases"
+        sx={{
+          color: "inherit",
+          textDecorationColor: "currentColor !important",
```
naps62, line 65: are we using `!important` already? 😕

gui/src/store/settings.ts
```diff
@@ -0,0 +1,39 @@
...
+
+interface State {
+  show: boolean;
+  actions: Action[];
+}
+
+interface Setters {
+  open: () => unknown;
+  close: () => unknown;
+}
```
naps62, line 14: when I saw a settings store, I expected something very different, equivalent to the other ones: to sync data from the settings that exist on the backend, and replace all `invoke` calls around the app

instead it's a store about the "settings window", not the settings themselves. is there a more clear name we can use?

bin/iron/src/app.rs
```diff
@@ -148,6 +225,9 @@ fn on_menu_event(event: WindowMenuEvent) {
         "close" => {
             event.window().close().unwrap();
         }
+        "settings" => {
+            event.window().emit("settings", ()).unwrap();
+        }
```
naps62, line 230: this is a very unclear name as well. `settings` is not an action. if I understood from context, `show-settings` would be better?

bin/iron/src/app.rs
```diff
@@ -109,6 +109,72 @@ impl IronApp {
...
+        menu = menu.add_submenu(Submenu::new("Edit", edit_menu));
+
+        // view submenu
+        #[cfg(target_os = "macos")]
+        {
+            menu = menu.add_submenu(Submenu::new(
+                "View",
+                Menu::new().add_native_item(MenuItem::EnterFullScreen),
+            ));
+        }
```
naps62, line 175: can this be broken down into other functions? perhaps moving to a separate file to clean things up

only reason I'm asking is because all the `cfg` attributes can easily lead to us messing the build for the OS we're not using (like I did in the past by accident).
it's very unclear here what exists where

bin/iron/src/app.rs
gabrielpoca: My reasoning is that the function is the handler for menu events `on_menu_event`, so "settings" is clicking the "Settings" button. It made sense, but I'm happy to change it.

bin/iron/src/app.rs
gabrielpoca: It was copy pasted from [tauri's source code](https://docs.rs/tauri-runtime/0.14.0/src/tauri_runtime/menu.rs.html#245), so I was trying to keep it close to the original, but I'll break it down 👍

bin/iron/src/app.rs
naps62: yes, and from that perspective it looks logic. but these are still global events, not a specific channel about menu events.


`
we could also do something different and namespace these to make it clear where they came from: `emit("menu:settings")`

gui/src/store/settings.ts
gabrielpoca: I understand, but I don't have great ideas. We do need state management for temporary things. What about calling it "window" store? This could evolve to control more pieces of information related to windows.

gui/src/components/NewVersionNotice.tsx
gabrielpoca: Unfortunately yes :/ I tried a lot of things. The `text-decoration-color` has a huge priority, and it's impossible to override without making a mess. This was the most harmless change I managed because I'm forcing the color of the underline to be whatever the current color is. I'm not even setting the color itself.

bin/iron/src/app.rs
gabrielpoca: Yap 👍 that makes sense

gui/src/store/settings.ts
naps62: > This could evolve to control more pieces of information related to windows.

is that the actual plan, or just a random idea?

if we don't really know how this will evolve, and what other data will be here, I'd rather just pick a more explicit name and rename it again later once we figure that out

`settings` to me is confusing because, from seeing all the other stores right now (balances, contracts, wallets, theme). They all store the thing described by the name. this one doesn't store the settings, but stores the "state of the setings window"

so something like "settingsWindow" works for me

"window" is also fine except that: 1) it's a reserved keyword already, and 2) means `open/close` need to be renamed to `openSettings / closedSettings` (which is fine to me)

bin/iron/src/app.rs
naps62: I'll handle this separately

gui/src/components/NewVersionNotice.tsx
naps62: apparently using `&&` worked https://dev.to/magischerzwerg/double-ampersand-trick-in-sass-with-react-4khe

Issue #294, opened 2023-07-12
Updating wallet settings does not refresh the initial state


naps62:
To reproduce:
1. Go to the test mnemonic settings
2. Toggle "is Dev Account"
3. Save
4. Toggle "is Dev Account" again
5. the form is now in a state where you can't save, because it matches the initial state, even though it was changed in be tween

expected:
the second save should be possible again

Also, arguably after saving we should maybe close the accordion automatically anyway

0xdorifto:
Hey guys, can I take this issue?

naps62:
Fixed in #337

PR #337, opened 2023-08-17
Edit wallet: able to edit wallets and re-edit to original state

0xdorifto:
Solution for https://github.com/iron-wallet/iron/issues/294

Before:

[Screencast from 17-08-2023 15:45:07.webm](https://github.com/iron-wallet/iron/assets/57015299/71b4cbf1-4c76-4974-b7e5-271cbe038d15)

After:

[Screencast from 17-08-2023 15:45:37.webm](https://github.com/iron-wallet/iron/assets/57015299/808ebc9e-9b60-4663-8427-c9fe27408c7f)


naps62:
Hey. thanks for the help!

Before testing this: I think the issue maybe wasn't fully clear now that I re-read. the problem is general about all settings. The checkbox was just the one example I used

in general: editing settings, and then editing back to what the original state was disables the button.

I'll probably merge this anyway as a quick fix for the checkbox, which is the most direct way to reach the error, but the issue won't be fixed by this

0xdorifto:
@naps62 yeah, I thought it was something specific to the checkbox. I'll keep looking at it then, no need to merge this.

0xdorifto:
[simplescreenrecorder-2023-08-29_19.56.49.webm](https://github.com/iron-wallet/iron/assets/57015299/bd7f804a-3de7-4318-8a39-337df86428f2)
This is how it is looking, with the save button closing the accordion.

0xdorifto:
Does it look good to you @naps62?

gui/src/components/Settings/Wallet/JsonKeystore.tsx
```diff
@@ -50,7 +53,8 @@ export function JsonKeystore({
           color="primary"
           variant="contained"
           type="submit"
-          disabled={!isDirty || !isValid}
+          disabled={(!isDirty || !isValid) && !isActuallyDirty}
+          onClick={() => setIsActuallyDirty(true)}
```
naps62, line 57: I'm not understanding this logic

clicking on the submit button makes the dirty state true? that sounds backwards. in fact, check this flow:

- edit an item (e.g.: the "dev account" checkbox)
- save the wallet, which closes the accordion
- open the wallet again. the save button is now enabled, even though it's in the original state

what I would imagine a correct solution does is:

- each wallet uses `defaultValues: wallet`. which is what's used to compare with current values and compute `isDirty`
- if we have this value updated (which should be happening automatically, but maybe there's some break along the stack I'm missing), this should "just work". Ideally we'd find out where the reactivity is missing and fix there

naps62: Thinking about this again, maybe its just a matter of calling https://react-hook-form.com/docs/useform/reset ?

0xdorifto:
@naps62 can you give me some feeback on this?

naps62:
Removed the auto-close accordion part, since I'm having second thoughts about it (and because the current logic was a bit limiting in other ways).

Also had to tweak the types a bit, but otherwise looks good

Issue #229, opened 2023-06-05
ws link being hit on every page I visit


xphoniex:
When I open the background.js inspector, I can see attempts like:

```
WebSocket connection to 'ws://localhost:9000/?origin=https%3A%2F%2Fcdn.coinzilla.io&tabId=736091678&url=https%3A%2F%2Fetherscan.io%2F...' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED
```

on basically any web page I visit. 

I was expecting such connections to only start after I connect the wallet and/or send an rpc request.

Is this normal or a bug?

PR #346, opened 2023-08-19
Lazy window.ethereum

naps62:
Previously, every browser tab created a new WS connection to the desktop wallet.
This resulted in unnecessary visual clutter in the UI, unnecessary resources being used (one WS connection being set up for every single browser tab), and possible confusion since the wallet appears to indicate random web2 pages are somehow connecting to the provider. This was first reported in #229 

The new behaviour does a few tricks to prevent this.
* Now, only upon the first `window.ethereum.request(...)` call is the connection initiated
* the background script, to account for the fact that ws handshake is async, keeps a backlog of incoming msgs, and flushes them in `Websockets.onOpen`.

Would be curious to get input from people more experienced in browser<->wallet connections. maybe there's some detail here I might be missing that could be a problem. Random thoughts:
* `window.ethereum.chainId` and other such fields will be undefined until this first request is triggered. wondering if this can somehow trip wallet SDKs
* can this delay cause any noticeable UI latency or flicker on page load?


naps62:
long shot: summoning @jxom here since their team may have insight into whether or not this is a good idea

jxom:
Makes sense to me. `window.ethereum` is just an implementation of an EIP-1193 Provider, and only requires `.request`, `.on` and `.removeEventListener` to be implemented. 

Other attributes such as `chainId`, etc aren’t necessary, and Dapps shouldn’t assume that these non-standard attributes exist. They should ideally be calling `eth_chainId` instead.

extension/src/inpage/provider.ts
```diff
@@ -244,13 +251,19 @@ export class IronProvider extends SafeEventEmitter {
    * if provided and marks this provider as initialized.
    * Throws if called more than once.
    */
-  protected async initializeStateAsync() {
+  protected async initialize() {
```
gabrielpoca, line 254: Is it possible to have concurrent calls to `initialize`? Asking because you're only setting `this.state.initialized` at the end of the function.

gabrielpoca:
I'm not sure if this would work, but you can maybe transform the `.chainId` into an object proxy method, and it calls the initialization function in its implementation.

extension/src/inpage/provider.ts
naps62: only setting at the end because it's fallible. need to make sure the request succeeds before assuming it's initialized (e.g.: if the app is not even running)

if you check all the `handle*` functions, they already avoid emitting duplicate events if the previous state is the same, so this should be safe

Issue #150, opened 2023-05-16
EIP 6963


naps62:
https://eips.ethereum.org/EIPS/eip-6963

PR #352, opened 2023-08-25
feat: eip-6963

jxom:
Implementing EIP-6963. 

Feel free to modify the `icon` – just using the GH org one lol.


naps62:
Awesome! I'll give this a test once I have a chance, but looks pretty straightforward

It will also close #150

naps62:
Just switched to an URI-encoded svg logo (seems to be mandatory according to the spec)

Issue #267, opened 2023-06-27
Random issues connecting tabs to desktop app


naps62:
As reported by @xphoenix in #228 

> Another issue I'm encountering is, if I'm logged in a dapp, on next refresh, it won't work. Sometimes I have to refresh a few times for it to reconnect to tauri again.
>
> I haven't tested on Chrome yet, and not sure if this is only a Firefox thing. Any similar experience?

I personally saw this happening a couple of times, but not yet in a reproducible way. need to investigate further

naps62:
Fixed in #356

PR #356, opened 2023-08-28
Fix a long-standing bug where cs <> bg communication would sometimes fail

naps62:
This one has been eluding me for a few weeks now, but should finally bring some more stability to the extension connection.

Seems to be a chromium bug first described here: https://groups.google.com/a/chromium.org/g/chromium-extensions/c/gHAEKspcdRY?pli=1

Metamask applies a similar workaround here: https://github.com/MetaMask/metamask-extension/blob/64aef2a1c9f5b46fdf2aa52f8d94750dc9873496/app/scripts/contentscript.js#L642-L652


naps62:
This also seems to be the solution to #267

PR #363, opened 2023-08-31
Extension build script now compatible with linux

0xdorifto:
Screenshot from 2023-08-31 10-35-02

extension build script wasn't working on linux. Fix and tested in Mac and Ubuntu.

### DIFF
--- a/extension/scripts/build.sh
+++ b/extension/scripts/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh -ue
+#!/bin/sh
 
 
 #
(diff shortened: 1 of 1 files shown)


naps62:
hmm. weird, I use Linux as well, with no problems (Arch & Manjaro). what distro are you running?

the fix seems fine either way, though

0xdorifto:
@naps62 I run Ubuntu.

naps62:
note (not sure if you can check @0xdorifto , but if not I can try myself later)

can we try changing the shebang to `#!/bin/bash` instead? if that also fixes it, I would prefer it over this
if not, I can merge as is

0xdorifto:
yeah, we can also do that and it works, I've tried it before

PR #365, opened 2023-09-01
Show a warning when an alchemy key is necessary for Iron to work properly

gabrielpoca:
In this change, we add a notice when the user uses a network supported by Alchemy but doesn't have an Alchemy key setup. This should make it straightforward for users that they do not see balances, txs, etc., because the key is not set.

Extra:
* Added the dependency notistack to handle stacking notices on top of each other. This also allowed the removal of the style overrides for Alerts in the theme.
* Because of notistack, I refactored the existing notice for new versions to be a hook instead of a component.
* Added a notification when the settings are changed so the UI can react to it.
* Added a command to check if a chain is supported by alchemy.

Screenshot 2023-09-01 at 16 55 43


gabrielpoca:
@naps62 ended up changing many things during the refactor, let me know what you think

bin/iron/src/app.rs
```diff
@@ -59,7 +59,8 @@ impl IronApp {
                 iron_forge::commands::foundry_get_abi,
                 iron_rpc::commands::rpc_send_transaction,
                 iron_connections::commands::connections_affinity_for,
-                iron_connections::commands::connections_set_affinity
+                iron_connections::commands::connections_set_affinity,
+                iron_sync::commands::alchemy_supported_network
```
naps62, line 63: can you rename to `sync_alchemy_supported_network`? to keep the convention of starting the command name with same name of the crate it's in

crates/sync/src/commands.rs
```diff
@@ -0,0 +1,4 @@
+#[tauri::command]
+pub async fn alchemy_supported_network(chain_id: u32) -> bool {
```
naps62, line 2: when I read the `app.rs` above I didn't realize this.
shouldn't this be `alchemy_supports_network` (or even `alchemy_is_network_supported` and rename the other function as well)

gui/src/components/Notices.tsx
naps62, line 1: empty file?

gui/src/components/Settings/General.tsx
```diff
@@ -42,12 +43,15 @@ export function SettingsGeneral() {
...
     async (data: FieldValues) => {
       reset(data);
+
       await invoke("settings_set", {
         newSettings: data,
       });
-      mutate();
+
+      reloadSettings();
+      reloadTheme();
```
naps62, line 52: isn't this redundant with the store listeners?

gui/src/hooks/useRefreshSettings.ts
```diff
@@ -0,0 +1,5 @@
+import { useEventListener } from "./useEventListener";
+
+export function useRefreshSettings(callback: () => unknown) {
+  useEventListener("settings-changed", callback);
```
naps62, line 4: we could just type type `useEventListener` with the union of all supported events, and remove all these `useRefresh***` indirection?

gabrielpoca: I didn't get what you mean

naps62: that instead of `useRefreshSettings`, `useRefreshTransaction`, and all these others, we can just use `useEventListener` directly

and we type it as:

```typescript
type Event = "settings-changed" | "transactions-updated" | ...;
useEventListener(event: Event, ...)`;
```

gui/src/components/Settings/General.tsx
gabrielpoca: 👍 I was missing a subscribe, updated

gui/src/hooks/useRefreshSettings.ts
gabrielpoca: I guess these listeners will always be the same 👍

gabrielpoca: done

gabrielpoca:
ty @naps62 🙏 I'm going to finish this up now that I'm back from vacations

gabrielpoca:
@naps62 I rebased the PR with main and did a couple of changes.

Now, about the discussion we had a few weeks ago, the issue with this is that we are checking for the existence of an alchemy key, but we don't know if the key is working. Sure, we can add a validation in the form to ensure the key works before we save it to storage, but if someone deletes their key in Alchemy's dashboard, we will never be informed that there's an error. I still think this change makes sense, but we may also need some general-purpose mechanism to notify the user of errors in the background, for instance, during sync.

naps62:
@gabrielpoca I agree. sounds like we also need to discuss how that would work though

I would focus on this individual notification for now, until we understand how best to handle these more generally on the frontend (maybe during the retreat we can discuss it?)

gabrielpoca:
@naps62 👍 shall we merge this then?

naps62 (approved):

PR #367, opened 2023-09-02
Fixes crash on wrong password

naps62:
Fixes a rust crash when the unlock password was incorrect.
The correct behaviour is for an `Err` to be returned to update the UI, but instead there was an `unwrap()` somewhere in there


naps62:
cc/ @gabrielpoca

gabrielpoca (approved):
works 👍

Issue #362, opened 2023-08-30
Dark mode selector not working


0xdorifto:
[darkmode-2023-08-30_11.30.35.webm](https://github.com/iron-wallet/iron/assets/57015299/b3af9c73-3ad5-4c5f-a7b2-66cc99b7d087)

It is not possible to get out of the automatic setting.

naps62:
hmm. it seems the setting is working, but the theme stopped reloading after saving settings.

if you right-click & reload, or if you do it via `Ctrl + K`, it works.
So it's still a bug, but just scoped to the settings form

PR #369, opened 2023-09-05
Fix dark mode selector not working

0xdorifto:
fixes this issue: https://github.com/iron-wallet/iron/issues/362


naps62:
hey 👋🏼 
Apparently the changes @gabrielpoca made in #365 already solve this, sort of by accident.
the main problem is that changes to `Settings` weren't being notified to the UI, so the store could reload. he ended up solving that during his PR.

Updating the theme specifically would only be a short-term solution, since it doesn't handle any other present or future settings that may affect the UI

Issue #371, opened 2023-09-06
move from IPC to HTTP


naps62:
## Problem

We currently rely on `#[tauri::command]` to define endpoints to communicate between the UI and the desktop app.
This heavily couples all of our components to tauri, and makes it difficult to work with other layers. e.g.: the app doesn't work on a browser (even though it is running on `localhost:1420`), largely because of this. It's also impossible to communicate with this endpoints from outside the JS in the webview.

## Solution

Create a new `iron_http`, to live side by side with `iron_ws`. the root endpoint should behave as an ethereum RPC, and an additional `/iron/` namespace should include endpoints for dealing with all the things that are currently tauri commands

the app should now launch both a WS and an HTTP server on boot. Both can be used as an Ethereum provider

## Checklist
- [x] Create `iron-http` crate #372 
- [x] Have root endpoint behave as an Ethereum RPC #377
- [ ] Add `/iron/` namespace endpoints for responding to tauri commands
  - [x] `forge_get_abi` https://github.com/iron-wallet/iron/pull/391
  - [x] `forge_get_name` https://github.com/iron-wallet/iron/pull/427
  - [x] `ws` https://github.com/iron-wallet/iron/pull/421
  - [x] `connections` #429
  - [x] `networks` #430 
  - [x] `settings` #431
  - [x] `wallets` #432
  - [x] `simulator` #433
  - [x] `db` #439

PR #372, opened 2023-09-07
Add iron-http crate with initial Axum setup

DavideSilva:
Why:
* Starting the migration proposed in issue #371

How:
* Creating a new crate in `crates/http` that spawns an Axum server.
  The server for now only responds to the `/` endpoint with a "Hello
  world!" message. The server runs by default at `127.0.0.1:9003` but
  can be configured through an env var `IRON_HTTP_SERVER_ENDPOINT`
* Adding the new crate to `Cargo.toml`
* Calling the `init` function from the `iron-http` crate in
  `bin/iron/src/app.rs`

### DIFF
--- a/crates/http/src/init.rs
+++ b/crates/http/src/init.rs
@@ -1,0 +1,17 @@
+use axum::{response::Html, routing::get, Router};
+
+pub async fn init() {
+    tokio::spawn(async {
+        let routes = Router::new().route("/", get(Html("Hello world!")));
+
+        let addr = std::env::var("IRON_HTTP_SERVER_ENDPOINT")
+            .unwrap_or("127.0.0.1:9003".into())
+            .parse()
+            .unwrap();
+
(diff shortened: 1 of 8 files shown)


crates/http/src/init.rs
```diff
@@ -0,0 +1,17 @@
+use axum::{response::Html, routing::get, Router};
+
+pub async fn init() {
+    tokio::spawn(async {
+        let routes = Router::new().route("/", get(Html("Hello world!")));
+
+        let addr = std::env::var("IRON_HTTP_SERVER_ENDPOINT")
```
naps62, line 7: can you also change the ws env name to `IRON_WS_...` to match?

naps62 (approved):

crates/http/src/init.rs
DavideSilva: Yup, not a problem

Issue #364, opened 2023-09-01
Frontend allowing invalid mnemonic


0xdorifto:
Frontend is allowing invalid mnemonic phrases which is breaking the app:

[1-2023-09-01_15.37.54.webm](https://github.com/iron-wallet/iron/assets/57015299/6953ad76-a4a9-4195-81bc-07100dafcc46)

naps62:
hmm, I thought I had fixed this a while ago, but maybe I'm misremembering. thanks for reporting

0xdorifto:
I can take this one.

naps62:
this was fixed in #374

PR #374, opened 2023-09-08
Refine mnemonic schema

0xdorifto:
Solution for this issue: https://github.com/iron-wallet/iron/issues/364

### DIFF
--- a/gui/src/types.ts
+++ b/gui/src/types.ts
@@ -1,3 +1,5 @@
+import { validateMnemonic } from "@scure/bip39";
+import { wordlist } from "@scure/bip39/wordlists/english";
 import { z } from "zod";
 
 export const generalSettingsSchema = z.object({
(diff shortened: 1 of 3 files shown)


gui/src/types.ts
```diff
@@ -66,7 +68,10 @@ export const mnemonicSchema = z
       message:
         "Invalid number of words. Needs to be 12, 15, 18, 21 or 24 words long",
     }
-  );
+  )
+  .refine((data) => validateMnemonic(data, wordlist), {
+    message: "This word combination is not valid",
+  });
```
naps62, line 74: hmm, this looks very much like what I would expect, but I was aiming to have this be done by the backend, instead of introducing extra dependencies on the UI

the reason being that the rust code is the source of truth, and the one that will ultimately parse the mnemonic. if there's some small inconsistency, it could lead to problems)

what I was thinking was adding a `iron_wallets::utils::validate_mnemonic` function, expose it as a command, and do:
```
refune(async (mnemonic) => await invoke("wallets_is_mnemonic_valid", { mnemonic })
```

I'll probably end up merging this as is and leaving that for future work, though

naps62 (approved):

Issue #224, opened 2023-06-03
Tx review: ability to tweak gas parameters


naps62:

0xdorifto:
Can I take this one?

0xdorifto:
@naps62 sounds good to me.

naps62:
Some random thoughts that may help:

* in terms of UI, this should allow tweaking the usual EIP1159 gas params
* probably keep this in a collapsed section of the popup, so that it doesn't get in the way unless people do want to tweak them
* when values are changed, the UI should trigger a simulation of the tx to display the usual feedback of whether the TX will succeed or not. this request doesn't need to be handled at first, but probably the UI should be built with that in mind

PR #376, opened 2023-09-08
Add gas customization UI

0xdorifto:
issue: https://github.com/iron-wallet/iron/issues/224

What do you think about this UI @naps62?

Screenshot from 2023-09-08 13-59-27

Screenshot from 2023-09-08 13-59-36


naps62:
Quick thoughts:
- what's the "gas" field? this doesn't seem EIP-1559 compliant
- should be filled with the default values coming from the backend
- "set gas" doesn't seem very intuitive. I'd expect some `onChange` logic like metamask does, to trigger a simulation in the background

PR #377, opened 2023-09-08
Correctly respond to JSON-RPC requests on `iron-http`

DavideSilva:
Why:
* Continuing the migration proposed in issue #371

How:
* Adding a `handler_rpc` function that handles JSON-RPC requests made to `/`,
  using `iron_rpc::Handler`
* Replacing the temporary implementation of the router in `http/init.rs` to make use
  of the new handler function
* Adding an `error.rs` with error type declarations
* Making the needed changes to `lib.rs` and `Cargo.toml`

With this change, it's now possible to use Iron as a JSON-RPC provider, via HTTP:

```js
import { ethers } from "ethers";

const provider = new ethers.JsonRpcProvider("http://127.0.0.1:9003");

console.log(await provider.getBlockNumber());
```

### DIFF
--- a/crates/http/src/error.rs
+++ b/crates/http/src/error.rs
@@ -20,8 +20,9 @@ impl serde::Serialize for Error {
     }
 }
 
+// TODO: revisit this to correctly respond with the appropriate status code
 impl IntoResponse for Error {
     fn into_response(self) -> Response {
-        (StatusCode::INTERNAL_SERVER_ERROR, "UNHANDLED_CLIENT_ERROR").into_response()
+        (StatusCode::INTERNAL_SERVER_ERROR, self.to_string()).into_response()
     }
 }
--- a/crates/http/src/error.rs
+++ b/crates/http/src/error.rs
@@ -1,0 +1,27 @@
+use axum::{
+    http::StatusCode,
+    response::{IntoResponse, Response},
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("serialization error: {0}")]
+    Serde(#[from] serde_json::Error),
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+impl serde::Serialize for Error {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.to_string().as_ref())
+    }
+}
+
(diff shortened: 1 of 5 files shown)


crates/http/src/error.rs
```diff
@@ -0,0 +1,27 @@
...
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.to_string().as_ref())
+    }
+}
+
+impl IntoResponse for Error {
+    fn into_response(self) -> Response {
+        (StatusCode::INTERNAL_SERVER_ERROR, "UNHANDLED_CLIENT_ERROR").into_response()
```
naps62, line 25: I believe this should also work, no?

```suggestion
        (StatusCode::INTERNAL_SERVER_ERROR, self.to_string()).into_response()
```

naps62: also, maybe work for later, but internal server error may not be correct.
e.g. if this was a serde_json error when parsing request data, it could mean the input was wrong, in which case this would be Unprocessable Entity.

I'm fine with keeping this as the only case though, if it makes it simpler

crates/http/src/init.rs
```diff
@@ -15,3 +19,22 @@ pub async fn init() {
             .unwrap();
     });
 }
+
+fn routes_rpc() -> Router {
```
naps62, line 23: name seems weird. how about `rpc_proxy`?

crates/http/src/init.rs
```diff
@@ -15,3 +19,22 @@ pub async fn init() {
...
+    Router::new().route("/", post(handler_rpc))
+}
+
+#[derive(Debug, Deserialize)]
+struct RpcParams {
+    domain: Option<String>,
+}
+
+async fn handler_rpc(Query(params): Query<RpcParams>, payload: String) -> Result<Json<Value>> {
```
naps62, line 32: same here. `rpc_handler` seems more intuitive

naps62 (approved):
awesome 🎊 

you should now be able to pull from main to get the build green

one minor detail: can you add a checklist on the original issue, linking all the related PRs, and listing roughly what the work left is? that way someone else may pick up smaller tasks later if applicable (or if you end up with no time to finish), and we can track the overall progress

crates/http/src/error.rs
DavideSilva: I'll add a note to revisit this later on as it also might need to change for the `/iron/` endpoints

PR #382, opened 2023-09-18
Po/nfts list frontend

jopedroliveira:
This PR implements a frontend for displaying the user's NFTs. 
It implements a new panel with responsive grid, that opens a modal when an item is clicked; and implements a store that fetches data every minute, or when backend emits an event. 

https://github.com/iron-wallet/iron/assets/6967377/5867bd68-7d3c-43dd-939a-338b1b5b5ad5


gui/src/components/Sidebar.tsx
```diff
@@ -31,6 +33,7 @@ export const TABS = [
     component: Account,
     icon: RequestQuoteSharpIcon,
   },
+  { path: "nfts", name: "NFTs", component: Nfts, icon: ImageIcon },
```
naps62, line 36: instead of a new sidebar item, I think NFTs should go under the Account page.

probably later with some better navigation inside it (when many items exist), but that's for later
keeping NFTs on a separate page when Account already handles native + ERC20 is very inconsistent

gui/src/components/Nfts.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </Grid>
+
+      <Modal
+        sx={{width: '80%'}}
+        open={detailsViewOpen}
+        onClose={() => {
+          setCurrentNftDetails(undefined);
+          setDetailsViewOpen(false);
+        }}
+      >
```
naps62, line 50: I'm gonna let @gabrielpoca review the modal part since I'm not sure if:
1. this is the approach we want to go with?
2. the current logic plays well with the rest of the UI? (e.g.: what if we trigger the "command bar -> open settings" option while an NFT modal is open?

gui/src/components/Sidebar.tsx
jopedroliveira: About this, agree with the consistency but I don't agree that we should place everything under "Account" and mix erc20 and erc721 together, because it might become messy.
What about changing the "account" panel to "balances" or "tokens" and keep two separated pages ?

gabrielpoca: I have no idea about where to put this. It makes sense to put it on the Account, but that would require some sort of navigation to change between ERC20 and ERC721, while at the same time, there's plenty of space on the sidebar for now.

gui/src/components/Nfts.tsx
gabrielpoca: For me, the Settings are a particular type of modal, like the Command Bar, that can be opened and closed from anywhere without messing up the app's navigation. I think this modal NFT should be a page (for instance, `/nfts/:id`) for now. Visually, it can be a modal on top of the NFT list, as long as the settings and command bar always show up on top, but making it a page makes it so we can link to it, refresh without losing state, and avoid future problems of having way too many places where modals can be open.

gui/src/components/Sidebar.tsx
naps62: The simples approach I was thinking of this previously was to eventually have a submenu below "Account", which could just be made with anchor links within the same page. Meaning we can either scroll down and see the NFTs, or still navigate through the sidebar. for me this is the best of both worlds

the "Accounts" tab was already called "Balances" before, and the reason for changing that was precisely because I didn't want it to be just about ERC20s, but about the account as a whole

naps62:
closing this for now, it's too outdated at this point and we'll need to probably revisit from scratch

PR #385, opened 2023-09-22
Fixing settings checkboxes

naps62:


naps62:
@gabrielpoca solution inspired by this: https://stackoverflow.com/questions/64042394/react-hook-form-and-material-ui-formcontrol

gabrielpoca (approved):
😅 if it works

PR #391, opened 2023-09-29
Implementing `/api/foundry_get_abi` endpoint

DavideSilva:
Why:
* Implementing the `/api/*` endpoints for the `iron-http` crate
* Issue #371

How:
* Adding a new Router to handle the `/api` namespace
* Implementing the `/api/foundry_get_abi` endpoint that calls the
  `foundry_get_abi` function from the `iron-forge` crate

With this, it is now possible to request the ABI of contracts running on
a foundry node by doing a `GET` request to `/api/foundry_get_abi` and
passing in, as query params, the address and chain_id


crates/http/src/init.rs
```diff
@@ -24,6 +30,25 @@ fn rpc_proxy() -> Router {
     Router::new().route("/", post(rpc_handler))
 }
 
+fn iron_namespace() -> Router {
+    Router::new().route("/iron/foundry_get_abi", get(foundry_get_abi_handler))
```
naps62, line 34: can we add a foundry namespace as well so this becomes `/iron/foundry/abi`? the "get" part is implicit in the HTTP metho

crates/http/src/error.rs
```diff
@@ -7,6 +7,9 @@ use axum::{
 pub enum Error {
     #[error("serialization error: {0}")]
     Serde(#[from] serde_json::Error),
+
+    #[error("foundry error")]
+    Foundry(String),
```
naps62, line 12: I think you should be able to do

```suggestion
    Foundry(#[from] iron_forge::Error),
```

DavideSilva: I don't think I can because the `foundry_get_abi` function in the`iron-forge` crate returns a String on the `Result<Option<Abi>, String>` and not an Error

naps62: ah. then that needs to be fixed on the other side. something for later then 👍🏼

naps62 (approved):

PR #401, opened 2023-10-02
HTTP API

naps62:


gui/src/hooks/useApi.ts
```diff
@@ -0,0 +1,25 @@
+import useSWR, { SWRResponse } from "swr";
+
+const BASE_URL = "http://localhost:9004/iron";
```
DavideSilva, line 3: Default port for the HTTP endpoint server is 9003. We can change it there if you'd prefer 9004 but we should be consistent with the port used

naps62: this needs to somehow come from env vars too here.
the reason this has 9004 is because I have custom env values. I used to have port 9003 for the websockets (so that I can have both the dev version and a release running at the same time)

DavideSilva: Agree we should prefer setting things using env vars.
However, the default values should also match, in case the env vars are not set, don't you agree?

naps62: I agree. but again, I was stating that this was using 9004 here because it's what I use locally. this is a draft PR only. this definitely needs to be fixed before merging

PR #421, opened 2023-10-06
Implement http endpoints for ws commands

DavideSilva:
Why:
* Continuing adding endpoints for issue #371

How:
* Updating the `router` to merge all the `iron` namespaced endpoints
  under `/iron`
* Adding a `get_peers_by_domain_handler` function that calls the
  `ws_peers_by_domain`
* Updating the `Cargo.toml` to include the `iron-ws` crate as dependency

### DIFF
--- a/crates/http/src/routes/ws.rs
+++ b/crates/http/src/routes/ws.rs
@@ -1,0 +1,10 @@
+use axum::Json;
+use serde_json::Value;
+
+use crate::Result;
+
+pub(crate) async fn get_peers_by_domain_handler() -> Result<Json<Value>> {
+    Ok(Json(serde_json::to_value(
+        &(iron_ws::commands::ws_peers_by_domain().await),
+    )?))
+}
--- a/crates/http/src/routes/ws.rs
+++ b/crates/http/src/routes/ws.rs
@@ -1,10 +1,10 @@
+use std::collections::HashMap;
+
 use axum::Json;
-use serde_json::Value;
+use iron_ws::peers::Peer;
 
 use crate::Result;
 
-pub(crate) async fn get_peers_by_domain_handler() -> Result<Json<Value>> {
-    Ok(Json(serde_json::to_value(
-        &(iron_ws::commands::ws_peers_by_domain().await),
-    )?))
+pub(crate) async fn get_peers_by_domain_handler() -> Result<Json<HashMap<String, Vec<Peer>>>> {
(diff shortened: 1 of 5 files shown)


crates/http/src/routes/ws.rs
```diff
@@ -0,0 +1,10 @@
+use axum::Json;
+use serde_json::Value;
+
+use crate::Result;
+
+pub(crate) async fn get_peers_by_domain_handler() -> Result<Json<Value>> {
+    Ok(Json(serde_json::to_value(
+        &(iron_ws::commands::ws_peers_by_domain().await),
+    )?))
```
naps62, line 9: why change the return value here instead of keeping the original `Vec<Peer>`? I would assume it serializes automatically

DavideSilva: Couple of reasons. First, the `peers` module in `crates/ws` is private, so we can't use the `Peer` struct without making it public to other crates. Not an issue, just a statement.

The other reason is that `HashMap` does not implement the `Handler` trait that Axum requires for functions that handle requests. It seemed simpler convert to JSON than to deal with implementing the trait

Doing this
```rust
pub(crate) async fn get_peers_by_domain_handler() -> Result<HashMap<String, Vec<Peer>>> {
    Ok(iron_ws::commands::ws_peers_by_domain().await)
}
```

results in the following error

```
the trait bound `fn() -> impl std::future::Future<Output = std::result::Result<std::collections::HashMap<std::string::String, std::vec::Vec<iron_ws::peers::Peer>>, error
::Error>> {routes::ws::get_peers_by_domain_handler}: axum::handler::Handler<_, _, _>` is not satisfied
the following other types implement trait `axum::handler::Handler<T, S, B>`:
  <axum::handler::Layered<L, H, T, S, B, B2> as axum::handler::Handler<T, S, B2>>
  <axum::routing::MethodRouter<S, B> as axum::handler::Handler<(), S, B>> [E0277]
```

Is this what you meant or did you had something else in mind?

naps62: You're missing the `Json(_)` part in that last example which I think is the reason why it fails to compile

DavideSilva: You're quite right, adding `Json(_)` solves the issue. I'm going to update this to keep the same return value

naps62 (approved):

PR #425, opened 2023-10-09
Fix TX simulation bug

DavideSilva:
Why:
* Doing a transaction was causing the TX simulation window to not open
  properly

How:
* Interpolating the address passed to the `truncateEthAddress` function
  to prevent JS to assume the address as an hexstring and not a string

Co-Authored-By: Resende <17102689+ZePedroResende@users.noreply.github.com>

### DIFF
--- a/gui/src/components/AddressView.tsx
+++ b/gui/src/components/AddressView.tsx
@@ -37,7 +37,7 @@ export function AddressView({ contextMenu, address, copyIcon }: Props) {
 
   const content = (
     <>
-      {alias ? alias : truncateEthAddress(address)}
+      {alias ? alias : truncateEthAddress(`${address}`)}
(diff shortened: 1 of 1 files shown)


naps62 (approved):

gui/src/components/AddressView.tsx
```diff
@@ -37,7 +37,7 @@ export function AddressView({ contextMenu, address, copyIcon }: Props) {
 
   const content = (
     <>
-      {alias ? alias : truncateEthAddress(address)}
+      {alias ? alias : truncateEthAddress(`${address}`)}
```
naps62, line 40:
```suggestion
      {alias ? alias : truncateEthAddress(address.toString())}
```

DavideSilva: This actually doesn't work. The only way to fix it seems to be with string interpolation

PR #431, opened 2023-10-12
Implement `/iron/settings/` endpoints

DavideSilva:
Why:
* Continuing adding endpoints for issue #371

How:
* Implementing endpoints for the `iron-settings` commands
* Updating the `iron-http` router to include the `/settings` routes

### DIFF
--- a/crates/http/src/routes/settings.rs
+++ b/crates/http/src/routes/settings.rs
@@ -1,0 +1,86 @@
+use axum::{
+    extract::Query,
+    routing::{get, post},
+    Json, Router,
+};
+use iron_settings::{DarkMode, SerializedSettings};
+use iron_types::ChecksummedAddress;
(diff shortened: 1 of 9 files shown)


crates/http/src/routes/settings.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct FastModeParams {
+    mode: bool,
+}
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct GetAliasParams {
+    address: ChecksummedAddress,
+}
```
naps62, line 31: I'm wondering. is there no way to avoid having these individual structs when it's just 1 or 2 fields?

feels extremely verbose

naps62 (approved):

PR #442, opened 2023-10-17
Using the HTTP API across the whole UI

naps62:


naps62:
Will close this for now. will keep the API limited until it is secured

PR #489, opened 2023-11-10
PGP-backed wallets

naps62:


pferreira101:
Writing this comment for visibility and traceability as I won't be able to finish the feature.

Some prints of the current state of the feature:

- Trying to unlock the `gpg` file in order to get the mnemonic:
Screenshot 2023-11-23 at 16 25 38

- Validate the file exists:
Screenshot 2023-11-23 at 16 23 26

- Inserted wrong passphrase:
Screenshot 2023-11-23 at 16 25 46

- File containing invalid mnemonic:
Screenshot 2023-11-23 at 16 24 33

- Second step of the creation form:
Screenshot 2023-11-23 at 16 26 18

- Reloaded `gpgagent` to clear cache and prompt passphrase again. Changing the derivation path asks for the passphrase again, as it is needed to read the secret to derivate the addresses, instead of keeping the mnemonic in memory:
Screenshot 2023-11-23 at 16 26 51


Some handover considerations:
- The second step of the form is almost an exact copy of the `HDWalletForm` review step. Should extract common logic and reuse it;
- At the moment, only the creation form is coded, the update wallet form hasn't been done yet;
- I'm on Mac, and the default `pinentry` program (`/usr/local/opt/pinentry/bin/pinentry` as per `gpgconf`) wasn't prompting on a built version of the app (`iron.app`, `iron.dmg`). In dev mode, this prompt would happen on the terminal, instead of using the system keychain dialog. Had to replace the default `pinentry` with this one: https://formulae.brew.sh/formula/pinentry-mac. Not sure  of what the behaviour would be on linux and windows;
- Updated `.github/workflows/rust.yml ` or else `clippy` would fail;
- I think the requirements at `CONTRIBUTING.md` need to be updated to include `gpgme`, I also needed to install `pkg-config` in order to be able to build the app. This should extend to the release workflow.
- `is_unlocked` and `unlock` methods are still not implemented;
- At the `http` crate, I'm not sure if `get_mnemonic_addresses_from_pgp` works. It seems like `cargo` is not checking the file.
- `Clippy` is tripping, or am I missing something(?): 
Screenshot 2023-11-23 at 16 54 42

Issue #283, opened 2023-07-02
use tokenlists.org for icon/address verification


naps62:
We want to build (at compile-time) a token list for multiple chains
We can fetch them from tokenlists.org (or from the lists they refer to, actually)

we should construct our own erc20s.json, grouped by chain_id, and including things like name, symbol, decimals, and other details (probably follow the same schema listed on tokenlists.org

## Random details:

- probably build this under a `crate/tokenlist/` crate
- create a `build.rs` script that should download one or more of those lists
- parse each list to our own wanted format, e.g.:

```
{
  chain_id: {
    "0x....": {
      "name": "USDC"
    },
    "0x....": {
      "name": "DAI"
    },
    ...
  }
}
```

For now I'd argue only chain_id, address and name are needed. In the future, we may want to add other fields (price oracle address, coingecko ID, bridging information, but for now this should suffice

oboutarf:
Hello i would like to be assigned on this issue!

naps62:
Go ahead @oboutarf ! thank you

naps62:
@oboutarf any updates?

PR #503, opened 2023-11-16
Add tokenlist crate and create list.json

faustoxvx:
This PR addresses the initial steps of this issue: https://github.com/iron-wallet/iron/issues/283

Why?

- As an extra security measure, Iron Wallet should show crypto icons from a verified list fetched from a reliable source. Currently the chosen source is: https://gateway.ipfs.io/ipns/tokens.uniswap.org

How?

- Creation of a tokenlist crate that fetches and parses the extensive information coming from the source
- Creation of a list.json file built at build time
- Parse the information in a format that benefits future implementations
- Add tokenlist to .gitignore

Currently, it creates the list.json file in the iron/bin directory, as shown in this screenshot:
Screenshot 2023-12-06 at 15 23 42

### DIFF
--- a/bin/iron/build.rs
+++ b/bin/iron/build.rs
@@ -1,3 +1,6 @@
 fn main() {
+    println!("cargo:rerun-if-changed=bin/iron/token_list.json");
+    iron_token_list::build();
+
     tauri_build::build();
 }
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ tsconfig.tsbuildinfo
 target/
 !migrations/*.sql
 gui/vite.config.ts.timestamp-*
+token_list.json
 
 # extension artifacts
 extension/*.zip
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1223,8 +1223,10 @@ checksum = "7f2c685bad3eb3d45a01354cedb7d5faa66194d1d58ba6e267a8de788f79db38"
 dependencies = [
  "android-tzdata",
  "iana-time-zone",
+ "js-sys",
  "num-traits",
  "serde",
+ "wasm-bindgen",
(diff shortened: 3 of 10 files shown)


bin/iron/src/app.rs
```diff
@@ -127,6 +127,11 @@ async fn init(app: &tauri::App, args: &Args) -> AppResult<()> {
     iron_networks::init(resource(app, "networks.json")).await;
     iron_forge::init().await?;
 
+    
+    if let Err(err)= tokenlist::init().await {
```
naps62, line 131: I don't think this should be here. as I said, the list should be built at compile time (in a `build.rs` script). don't think we need to init something here at runtime

crates/tokenlist/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    //symbol: String,
+    //decimals: i32,
+
+    #[serde(rename = "logoURI")]
+    logo_uri: String,
+}
+
+pub async fn init() -> Result<(), Error> {
+
+    let response = match reqwest::get("https://gateway.ipfs.io/ipns/tokens.uniswap.org").await {
```
naps62, line 32: the idea of the local timestamp is to avoid doing this request if the timestamp is too recent. so the flow should actually be to read the current timestamp first, if any

DavideSilva (approved):

bin/iron/build.rs
```diff
@@ -1,3 +1,6 @@
 fn main() {
+    println!("cargo:rerun-if-changed=bin/iron/token_list.json");
+    iron_token_list::build();
```
naps62, line 3: this one was suposed to be the build script of the token_list crate itself, not called from here

Issue #500, opened 2023-11-15
Keyboard shorctuts


naps62:
We already have a few hardcoded shortcuts:
- `Ctrl-1`, `Ctrl-2`, ..., `Ctrl-4` cycle through the side bar menus
- `Ctrk-k` opens the command bar

We now want to have more available shortcuts, and to have them all configurable through the app's settings.

## Functionality

In addition to the above shortcuts, here's a list of some extra ones we may want:

- switching between wallets (probably through number keys somehow?)
- switching between the sidebar menus (I believe you can already Ctrl-1, Ctrl-2, etc)
- open/close settings
- toggle fast mode

One potentially tricky detail here is that we need to distinguish between macOS and Windows/Linux, since one has Cmd and the others have Ctrl

## UI

What I'm thinking is a UI along these lines (taken from Obsidian, a markdown note-taking app):

![image](https://github.com/iron-wallet/iron/assets/283819/becdfef2-a88f-48f3-824c-eb0a544ec067)

naps62:
don't think I'll be looking at this anytime soon. it honestly needs some validation on whether it is actually used

if revisiting in the future, I'd consider instead going with a vimium-like approach, so that we can achieve full keyboard navigation globally and in a less opinionated way

PR #504, opened 2023-11-16
Shortcuts

cstrecht:
This closes the Issue [#500](https://github.com/iron-wallet/iron/issues/500)

### Why: 
- This PR adds a new feature to the Iron Wallet: Navigation by using Shortcuts - Implementation of shortcuts functionality and creation of a new page on the Settings menu presenting a list with all the available shortcuts. 

### This change addresses the need by: 
👉 Creation of a list of keybinds. The user can access this list once the `Settings` menu is open. The keybinds that are currently working are: 

- `Ctrl + K` : Open and close the command bar;
- `Ctrl + [1..4]` : Switch between sidebar tabs;
- `Ctrl + S` : Opens and closes the Settings menu;
- `Ctrl + F` : Toggles Fast mode;
- `W` : Open the command bar already pre-filled with the option "Change wallet". The options are numbered, so the user can, for example, press `W` -> `2` -> `Enter`, and it switches to the second wallet available;
- `N`: Open the command bar already pre-filled with the option "Change network";
- `T`: Open the command bar already pre-filled with the option "Change theme".

 
 👉 A search bar for the `Keybinds` list that searches by keybind description.
 👉 A very simple and responsive UI. 
 
 ### Demo 👀

https://github.com/iron-wallet/iron/assets/77847398/eda51250-c55a-4907-8d89-4a71ba7b3aec

### Disclaimer: 
As I was developing the feature, I came to the conclusion that the usage of the [kbar library](https://kbar.vercel.app/docs/getting-started) gets to a point that it's not scalable. The shortcuts are working properly but if the user adds, for example, a new plaintext wallet, the kbar will not update unless the user reloads the page. This happens because we are statically adding the actions and for us to add a dynamic action that reacts on an user action (like creating a new wallet) requires us to use the `useRegisterAction` hook that, being a hook, must be used directly in a body of a component or inside another hook. 
This can be resolved in the future using the [cmdk library](https://cmdk.paco.me/) which better complies with React patterns, and will support creating new command bar items on user actions; or forking the kbar library and create a new way to dynamically add new actions without the need of an hook.


gui/src/components/Settings/Keybinds.tsx
```diff
@@ -0,0 +1,110 @@
...
+  ListItemText,
+  Stack,
+  TextField,
+  Typography,
+} from "@mui/material";
+import { useState } from "react";
+
+import { useWallets } from "@/store";
+
+const MAX_CHANGE_WALLET_SHORTCUTS = 9;
```
cstrecht, line 13: I had to redeclare the `MAX_CHANGE_WALLET_SHORTCUTS` constant twice due to linting issues.

Initially, when I created and exported the constant for use in the UI (`Sidebar.tsx`), the `yarn lint` failed due to import sorting. After fixing the sorting, the `yarn lint` passed, but running prettier caused a re-sorting, leading to a persistent loop. To resolve this, I opted for the current approach.

If you have any suggestions, please feel free to share.

gui/src/components/Sidebar.tsx
```diff
@@ -102,29 +151,11 @@ export function SidebarLayout({ children }: { children: ReactNode }) {
 
 export function Sidebar() {
   const [_match, params] = useRoute("/:path");
-  const [_location, setLocation] = useLocation();
+  const [_location] = useLocation();
```
DavideSilva, line 154: Is this used anywhere? If not, we can remove it

gui/src/components/Sidebar.tsx
```diff
@@ -65,24 +67,71 @@ export function SidebarLayout({ children }: { children: ReactNode }) {
...
-
   useKeyPress(
     range(1, TABS.length + 1).map(toString),
     { ctrl: true },
     handleKeyboardNavigation,
   );
 
+  //------------------------------------------------- Switch between wallets (Ctrl + Shift + number)
+  useKeyPress(
+    range(1, Math.min(wallets.length + 1, MAX_CHANGE_WALLET_SHORTCUTS)).map(
```
DavideSilva, line 113: What is the purpose of the call to `map(toString)`?

DavideSilva (changes requested):
Using the shortcut to switch wallet to number 3, CMD+Shift+3, triggers the MacOS screenshot feature

Also, the "Change Wallet" spotlight action is missing

gui/src/components/Sidebar.tsx
cstrecht: I'm using the `map(toString)` because the custom hook `useKeyPress` expects its first argument to be an array of strings. If I didn't call the `map(toString)`, the `range()` function would yield an array of numbers

gui/src/components/Settings/Keybinds.tsx
cstrecht: Update: this problem is already fixed.

cstrecht:
_"Also, the "Change Wallet" spotlight action is missing"_

Regarding the "Change Wallet" action being missed, I've managed to fix it.
Basically, once we're using [kbar](https://kbar.vercel.app/docs/concepts/actions) I should've kept the `useRegisterActions` hook because this hook is responsible for keeping dynamic register actions. It is now being used again

gui/src/components/Sidebar.tsx
paulacarneiro: A good alternative here would be to change the type of the first argument of `useKeyPress` to be an array of string OR an array of numbers, so we don't have to convert to string whenever we use number ranges. Just make sure to convert the keys of the array to string when comparing them to the event key

paulacarneiro:
> Using the shortcut to switch wallet to number 3, CMD+Shift+3, triggers the MacOS screenshot feature

I see you changed this shortcut to `ctrl + alt + [1..n]`, don't forget to update the keybind 'description' in the app settings and in the PR description. Also, are you planning on adding a `meta + alt + [1..n]` version of this shortcut? Just asking because the other shortcuts have both `ctrl` and `meta` versions, and if that's the case you will need to change the `useKeyPress` hook to check the event code instead of the key, as `meta+alt` modifies the key value

cstrecht:
Hey @paulacarneiro 

Thank you for your comment! Regarding your question about the `meta + alt + [1..n]`, I am developing a different approach for the "Change wallet" shortcut. The previous shortcut would not be scalable; I would need an arbitrary combination for switching networks, and then another one for switching between addresses inside each wallet. This could get confusing very easily.

For the "Change between wallet" shortcut, a clean and scalable solution would be a single shortcut, such as `ctrl+w`, which opens the command bar but is already prefilled with the "Change wallet" text. Then, each option in the command bar is numbered, so the user can simply do, for example, `ctrl+w`+`3`+`Enter` (and then choose the address). Naps helped me with this suggestion. I am currently working on implementing it. 🙌

gui/src/store/settings.ts
```diff
@@ -5,51 +5,39 @@ import { subscribeWithSelector } from "zustand/middleware";
...
+      subtitle: "enable/disable",
+      shortcut: ["˄ ", "+", " F"],
+    },
+    ...(["Enable", "Disable"] as const).map((mode, index) => ({
+      id: `${actionId}/${mode}`,
+      name: `${index + 1}: ${mode}`,
+      parent: actionId,
+      perform: () => {
+        const currentMode = mode === "Disable" ? false : true;
+        invoke("settings_set_fast_mode", { mode: currentMode });
```
paulacarneiro, line 32:
```suggestion
        invoke("settings_set_fast_mode", { mode: mode == "Enable" });
```

You don't really a ternary expression here since `a === b` already returns a boolean 😄

DavideSilva:
image

Aren't a couple of shortcuts missing from this list? 🤔 
The `T` and `N` shortcuts are not here. Should they be?

DavideSilva (approved):
Looks good :shipit:

naps62:
the Keybinds page suggests the shortcuts are uppercase, but they really aren't.
e.g.: `"W"` doesn't work, at least for me, but `w` does

PR #525, opened 2023-11-20
Yarn -> Bun

naps62:


naps62:
I'll leave this open as a draft for now. not sure if we want to go forward with ti

PR #527, opened 2023-11-21
Fix typo on general settings

pferreira101:
[
Screenshot 2023-11-21 at 12 32 19
](url)

Should be "minimized".


naps62 (approved):
Oh wow. I did that? 😅

PR #579, opened 2024-01-15
Fix Justfile dev command

DavideSilva:
Why:
* The recent change to the way Tauri is installed and managed broke the
  Justfile dev command

How:
* Using `yarn tauri dev` instead of `cargo tauri dev` to run the `dev`
  command

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -13,7 +13,7 @@ build:
   cargo build
 
 dev *args='':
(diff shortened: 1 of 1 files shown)


ZePedroResende (approved):

naps62:
I don't understand this one.
the recent changes to tauri are that now we're supposed to run `cargo dev`, not `yarn`. I made this change a few weeks ago

PR #583, opened 2024-01-22
Add new `tokenlist` endpoint

cstrecht:
## Why:
The `token-list` crate was not accessible by an endpoint from anywhere outside the project. This PR implements the `tokens` http endpoint. Given the pair `chainId` and `address`, the endpoint should return the associated token.

## This change addresses the need by:

Creation of a new `tokens` route group and `tokens/:chainId/:address` endpoint with `chainId` and `address` as path variables.

### Succeeds with a valid `chainId`/`address` pair : 
```bash
~ curl http://localhost:9003/iron/tokens/1/0x111111111117dC0aa78b770fA6A738034120C302
{
   "chainId":1,
   "address":"0x111111111117dC0aa78b770fA6A738034120C302",
   "name":"1inch",
   "logoURI":"https://assets.coingecko.com/coins/images/13469/thumb/1inch-token.png?1608803028"
}
```

### Errors with an invalid `chainId`/`address` data types:

```bash
~ curl http://localhost:9003/iron/tokens/a/0x111111111117dC0aa78b770fA6A738034120C302
Invalid URL: Cannot parse `chainId` with value `"a"` to a `i32`
```
### Errors with an invalid `chainId`/`address` pair:
```bash
~ curl http://localhost:9003/iron/tokens/1/0x1234567890                     
Not found
```


naps62:
I'm confused. was there an actual need for this endpoint? the point of generating the json is to embed it directly in the frontend eventually

cstrecht:
I do understand where you're coming from and it does indeed look like a better solution. To be honest, I didn't even think of that possibility, I was suggested to go this way (cc @DavideSilva wdyt about @naps62 approach?). 

This PR was my introduction to Rust and Axum, but of course we can go that route and only do this `token-list` changes on the frontend side of the codebase to use the json directly. Thx for the feedback!

DavideSilva:
@naps62 there's no actual need for this endpoint. This was made primarily as a learning exercise to learn about Axum and how to make an HTTP server endpoint

We were thinking this still might be useful if, for some reason, we want to expose an endpoint for users to query about a specific address and get the logo back, using Iron as a sort of wrapper

Is this something you'd see Iron doing in the future?

naps62:
Having the logos as part of the payload is also something I wouldn't do (at least for now). in fact, I removed them in my PR (#589 ).

the reason being that the logos being returned from the generated list were not suitable to show in the frontend. also, having a single logo URL is too simplistic, as we need to consider multiple sizes, and more importantly dark/light mode.
the frontend already has a list of icons to display. the point of the JSON list is purely to check if a token called "USDT" is actually the real USDT and not some scam coin with the same name

as for having this in the future: yeah, maybe. but it's still very speculative, and not hard at all to implement once that need does come

PS: I'm fine with the idea of using these tasks as an excuse to introduce concepts or other crates. but then we should be OK with not merging them

DavideSilva:
We are completely ok with not merging this, just thought it would be an interesting feature. As you said, this can easily be recovered or reimplemented on a later, if the need arises

Going to close this now

PR #585, opened 2024-01-23
Remove automatic errors from tokenlist

cstrecht:
wip


naps62:
I didn't notce this one before opening #589. but this solution wasn't the way to go anyway

abusing `map_err` and adding individual errors to the enum for every single possible error case is not a scalable approach. especially since this is build-time code, where whoever is reading it is the developer, who should be able to figure out details from context.

the original approach (using `#[from]`) is actually the correct approach most of the time, but I simplified it further by using `anyhow`, which just propagates the original error upwards

cstrecht:
Yep, I did felt that creating a new item on the enum for each new error could not be the best idea, but at the same time, having those specific errors covering the scope of a create didn't seem ideal either. `anyhow` solution looks better for me too, I was actually trying to find something like this initially. Do you want me to open a PR trying to do this?

PR #587, opened 2024-01-26
Return contracts with names

DavideSilva:


crates/db/src/commands.rs
```diff
@@ -52,6 +52,14 @@ pub async fn db_get_contracts(chain_id: u32, db: tauri::State<'_, DB>) -> Result
     db.get_contracts(chain_id).await
 }
 
+#[tauri::command]
+pub async fn db_get_contracts_with_name(
```
gabrielpoca, line 56: Do we really need to specify the "name" part? What about just using the previous function?

naps62:
Can you clarify what the purpose of this is?

DavideSilva:
In order to filter the contracts in the Contracts tab, we need to bring the fetching of the contract name up one level in the React component hierarchy. Right now, the `Contract` component receives each contract address and fetches the name to display. If we want to filter the contract list by name, we need to have the contract addresses and names in the `Contracts` component

crates/db/src/commands.rs
DavideSilva: I'll have to double check but I think the other function was used somewhere else and changing it broke something. I'll let you know after I look into this again

naps62:
Superseded by #616

PR #597, opened 2024-02-20
Updating to tauri 1.6

naps62:


naps62:
Currently blocked because it introduces dependencies that use BSL license (detected by `cargo-deny`)

naps62:
turns out this BSL is Boost Source License, not Business Software License, so we're good

Issue #285, opened 2023-07-03
Create denylist to hide spam/scam tokens


naps62:

naps62:
We already have a black list. but would be useful to be able to view it somewhere (in the settings, or hidden in the balances tab perhaps) so we could undo some if necessary

PR #696, opened 2024-07-03
Show/hide token feature (Issue #285)

josevazf:
This PR updates the Show/Hide token feature (Issue #285) by creating a new panel called Denylist under Settings where the hidden tokens are displayed, with the option to make them visible again under the Account tab.

**Why:**

- At the moment the tokens could only be hidden but there was no option to view and display them again.

**How:**

- Added new components to create the Denylist panel, display the tokens, and enable the option to show them back under the Account tab.
`gui/src/components/Settings/Denylist.tsx`
`gui/src/components/ERC20ViewDenylist.tsx`
`gui/src/components/Settings/Settings.tsx`
`gui/src/store/balances.ts`
`crates/http/src/routes/db.rs`

- Added new queries and commands to interact/update the database.
`bin/src/app.rs`
`crates/db/src/queries/erc20.rs`
`crates/db/src/commands.rs`

**Preview:**
image
image

**Other considerations:**

- Native chain token should not have the option to be hidden.
- With this update the Send button is disabled under the Denylist tab, to transfer the tokens they have to be visible under the Account tab.


naps62:
Overall love the update, but there's one change I'd like to make

it seems very weird to me to have the concept of a "denylist balance". they're only meant to show up in settings, so refering to balances there seems way off.
The fact that you're reusing the same ERC20View in that settings page also feels off. you're creating some tight coupling between the Balances list and the Denylist. Meaning whenever we change functionality for balances, we now have to worry how that'll affect the denylist (and most likely, we'll forget about it and cause weird behaviour)

What I'd do instead:
- just have a simple `db_get_erc20_denylist` call that lists out all the token addresses that have been hidden
- in the settings page, just show a simpler list element with all those (token name, address, but no balance), and a button to restore it

josevazf:
> Overall love the update, but there's one change I'd like to make
> 
> it seems very weird to me to have the concept of a "denylist balance". they're only meant to show up in settings, so refering to balances there seems way off. The fact that you're reusing the same ERC20View in that settings page also feels off. you're creating some tight coupling between the Balances list and the Denylist. Meaning whenever we change functionality for balances, we now have to worry how that'll affect the denylist (and most likely, we'll forget about it and cause weird behaviour)
> 
> What I'd do instead:
> 
> - just have a simple `db_get_erc20_denylist` call that lists out all the token addresses that have been hidden
> 
> - in the settings page, just show a simpler list element with all those (token name, address, but no balance), and a button to restore it

Thanks for the feedback! I will update accordingly 👍🏻

josevazf:
**Update:**

- Removed balance, simplified the list element and added a single button to restore the token (we could use the opposite button on the Account tab to hide the token, unless we are willing to have more options available)
- Fixed Denylist to be independent from the Balances definition

**Preview:**

image

bin/src/app.rs
```diff
@@ -48,7 +48,9 @@ impl EthUIApp {
                 ethui_db::commands::db_get_contract_abi,
                 ethui_db::commands::db_get_erc20_metadata,
                 ethui_db::commands::db_get_erc20_balances,
+                ethui_db::commands::db_get_erc20_denylist,
                 ethui_db::commands::db_set_erc20_blacklist,
+                ethui_db::commands::db_set_erc20_allowlist,
```
naps62, line 53: it's weird to call this allowlist because there actually isn't one. what this does is remove from the blacklist.

instead, I'd rather have `set_erc20_blacklist` and `clear_erc20_blacklist`

crates/db/src/queries/erc20.rs
```diff
@@ -146,6 +146,58 @@ impl DbInner {
             .collect())
     }
 
+    pub async fn get_erc20_denylist_balances(
```
naps62, line 149: based on my previous feedback, this function should no longer be needed right?

naps62:
Note: whenever you add/change SQL queries in the codebase, you'll need to run `mise run sqlx` to update the sqlx cache files

bin/src/app.rs
```diff
@@ -48,7 +48,9 @@ impl EthUIApp {
                 ethui_db::commands::db_get_contract_abi,
                 ethui_db::commands::db_get_erc20_metadata,
                 ethui_db::commands::db_get_erc20_balances,
+                ethui_db::commands::db_get_erc20_denylist,
```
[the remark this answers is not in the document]
josevazf: Yes, I wanted to confirm this with you to update all the `blacklist` references to `denylist`

bin/src/app.rs
josevazf: Agree!

crates/db/src/queries/erc20.rs
```diff
@@ -146,6 +146,57 @@ impl DbInner {
             .collect())
     }
 
+    pub async fn get_erc20_blacklist(
+        &self,
+        chain_id: u32,
+        address: Address,
+        include_blacklisted: bool,
```
naps62, line 153: I dont' understand this function. from the name, sounds like we're asking for the list of blacklisted tokens

so what is the `include_blacklisted` arg doing?

gui/src/components/Settings/Settings.tsx
```diff
@@ -14,13 +14,15 @@ import { SettingsGeneral } from "./General";
...
 import { SettingsWallets } from "./Wallets";
+import { SettingsBlacklist } from "./Blacklist";
 
 const TABS = [
   { name: "General", component: SettingsGeneral },
   { name: "Wallets", component: SettingsWallets },
   { name: "Network", component: SettingsNetwork },
   { name: "Foundry", component: SettingsFoundry },
   { name: "Keybinds", component: SettingsKeybinds },
+  { name: "Tokens", component: SettingsBlacklist },
```
naps62, line 25: Probably rename the component to `SettingsTokens` as well?

crates/db/src/queries/erc20.rs
josevazf: Yes it was redundant. I have updated!

naps62 (approved):
my apologies, this got lost in my backlog, last weeks have been a bit chaotic

seems good enough now. maybe a couple nitpicks, but nothing really important

Issue #695, opened 2024-06-28
Ability to render numbers in different formats


naps62:
Just like we already have an `AddressView` to add capabilities to address text (alias, copy to clipboard, etc), we should do the same for number and hash values that we want to display (e.g.: eth balances, transaction hashes, function call results, ...)

ideally with the following options:
- copy to clipboard
- convert between units (wei, eth, gwei, etc). only applicable to numbers, not hashes
- change format (decimal, hex)

naps62:
going to close this for now as I don't have a clear idea of how this should work anymore, or if we even want this right now

PR #697, opened 2024-07-04
Add NumberView and HashView components

helenamartins81:
**Why:** 

- Currently, it is not possible to convert numbers into different units (wei, gwei, eth) or different formats (decimal to hex)  (Issue #695) 

**How:**
- Added the NumberView and HashView components to enable conversion between units in numbers, copy to clipboard functionality, and format switching between decimal and hex.  
gui/src/components/NumberView.tsx
gui/src/components/HashView.tsx

**Preview**:
image

### DIFF
--- a/gui/src/components/NumberView.tsx
+++ b/gui/src/components/NumberView.tsx
@@ -1,4 +1,4 @@
-import { Grid, IconButton, Popover, Stack, Tooltip } from "@mui/material";
+import { Grid, IconButton, Popover, Stack } from "@mui/material";
 import { MoreVert as MoreVertIcon } from "@mui/icons-material";
 import { useEffect, useState } from "react";
 import { formatUnits, parseUnits } from "viem";
@@ -12,7 +12,9 @@ interface Props {
 }
 
 export function NumberView({ value, unit }: Props) {
-  const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null);
+  const [anchorEl, setAnchorEl] = useState<{
+    [key: string]: HTMLElement | null;
+  }>({});
   const [unitValuesOpen, setUnitValuesOpen] = useState(false);
   const [conversionResults, setConversionResults] = useState<{
     [key: string]: string | bigint;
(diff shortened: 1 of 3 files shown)


gui/src/routes/_home.home/transactions.lazy.tsx
```diff
@@ -223,24 +219,40 @@ function Details({ tx, chainId }: DetailsProps) {
         <>
           <Datapoint
             label="maxFeePerGas"
-            value={`${fullTx.maxFeePerGas && formatGwei(BigInt(fullTx.maxFeePerGas))} gwei`}
+            value={
+              fullTx.maxFeePerGas ? (
+                <NumberView value={Number(fullTx.maxFeePerGas)} />
```
naps62, line 224: this seems incorrect. `maxFeePerGas` and `maxPriorityFeePerGas` are already expressed in `gwei`, but that information is lost here

I'd expect something along the lines of

`<NumberView value={fullTx.maxFeePerGas} unit="gwei" />`

where the component would be smart enough to show me the conversions from there

gui/src/components/NumberView.tsx
```diff
@@ -0,0 +1,85 @@
...
+    } else if (formatValue == "Hexadecimal") {
+      setFormatValue("Decimal");
+      const numberValue = Number(value);
+      const hexadecimalValue = numberValue.toString(16).toUpperCase();
+      setDisplayValue("0x" + hexadecimalValue);
+    } else {
+      setDisplayValue(value.toString());
+    }
+    setMenuAnchor(null);
+  };
```
naps62, line 51: a "Change Format" button on the UI is very unclear as to what its purpose is.
from seeing the code, seems to toggle between hex and decimal notation. but for a user that won't be clear at all

naps62:
Overall a nice effort, but I feel this somewhat misses the point.

we can't blindly wrap values in a `NumberView` and offer a couple of static conversions. Some values already come in wei, others come in gwei or eth. We need to be smart about the parameters, and how to convert between them (viem already has functions to deal with all this, as long as we know the decimals)

What I was expecting, was not a context menu, but a modal showing the different values the number can take, so I can copy to clipboard the one I'm interested in

helenamartins81:
**Update:**

Removed the menu and added a modal with all the conversions (hashes only display hex and decimal values).
In NumbersView, adjusted the conversion process to consider the initial value unit.

**Preview:**

image

gui/src/components/NumberView.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export function NumberView({ value, unit }: Props) {
+  const [unitValues, setUnitValues] = useState({
+    wei: "",
+    gwei: "",
+    eth: "",
+    decimal: "",
+    hex: "",
+  });
```
naps62, line 22: wei/gwei/eth are not in the same group as decimal/hex. these are different axis
e.g.: I may want to view the same gwei value in decimal or hex

also, there are many more units other than these 3: https://www.gemini.com/cryptopedia/satoshi-value-gwei-to-ether-to-wei-converter-eth-gwei#section-ethereum-denominations-ether-to-wei-gwei-to-ether-more
for each of these, you can just map it to a number of decimal places, and use `formatUnits` from viem instead of manually handling each of them

helenamartins81:
**Update**:

In NumbersView, moved the hexadecimal format to another popover. Now, it is possible to see and copy the conversion to hexadecimal for each unit. I also added more units and adjusted the conversion process using formatUnits from viem.

**Preview:**

image

PR #698, opened 2024-07-16
Initial tour guide

helenamartins81:
**Why:**
Guided tour used to explain the purpose of some features during the onboarding;

**How:**
Using [react-joyride](https://docs.react-joyride.com/)

**Preview:**

image
image


naps62:
Overall this looks much better than the previous attempt.
Consider this approved, as far as main structure and decisions, but I'll refrain from merging right away since we need to properly figure out the copy & design

Would it be easy to somehow feature-flag this, so we could merge the code, but with the behaviour disabled behind an environment variable? So that we could iterate and test this without affecting future releases

Issue #152, opened 2023-05-16
EIP-747 - wallet_watchAsset


naps62:
https://eips.ethereum.org/EIPS/eip-747

PR #699, opened 2024-07-18
Alchemy Client refactor (Issue #152 - EIP747)

josevazf:
This PR servers as a needed update to the Alchemy Client to better enable the subsequent implementation of EIP-747.

**Why:**

- At the moment we have a fixed base URL defined under `crates/sync/alchemy/src/networks.rs`. To implement EIP-747, and possibly other proposals down the line, we need a way to dynamically update this URL based on the desired request.
As an example, to fetch an ERC-721 metadata we need to discard the existing `v2/` and replace it with `nft/v3/`.

**How:**

- Refactored the Alchemy Client to accept a string argument to dynamically update the URL as needed:
`crates/sync/alchemy/src/client.rs`
`crates/sync/alchemy/src/lib.rs`

- Functions using this Client now have to create a new instance where they pass the desired string to be added
`crates/sync/alchemy/src/client.rs`

### DIFF
--- a/crates/sync/alchemy/src/client.rs
+++ b/crates/sync/alchemy/src/client.rs
@@ -6,11 +6,16 @@ use ethers::providers::{
 use ethui_types::{events::Tx, Address, ToAlloy, ToEthers, TokenMetadata, U256, U64};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use url::Url;
 
 use crate::{networks, types::AlchemyAssetTransfer, Result};
 
 #[derive(Debug)]
-pub(crate) struct Client(Provider<RetryClient<Http>>);
+pub(crate) struct Client {
+    provider: Provider<RetryClient<Http>>,
+    chain_id: u32,
+    api_key: String,
+}
 
 pub(crate) enum Direction {
     From(Address),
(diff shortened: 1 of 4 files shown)


crates/sync/alchemy/src/client.rs
```diff
@@ -10,32 +10,45 @@ use serde_json::json;
...
+        Ok(Self {
+            provider,
+            chain_id,
+            api_key: api_key.to_string(),
+        })
     }
 
     pub async fn get_block_number(&self) -> Result<U64> {
-        let block = self.0.get_block_number().await?;
+        let client = Client::new(self.chain_id, &self.api_key, "v2/");
```
naps62, line 49: something's off here. You already have `self` which is a `client`, and you're now instantiating a new `Client` only to give it a different path

I understand you need to, in some circustances swap that last part of the URL, but this isn't the way to do it

you can have instantiate two providers on `new`, one regular for the previous behaviour, and an `nft_v3_provider` for the separate endpoint, and then use it where needed

naps62, line 49: something's off here. You already have `self` which is a `client`, and you're now instantiating a new `Client` only to give it a different path

I understand you need to, in some circustances swap that last part of the URL, but this isn't the way to do it

you can have instantiate two providers on `new`, one regular for the previous behaviour, and an `nft_v3_provider` for the separate endpoint, and then use it where needed

naps62 (changes requested):

naps62 (approved):

PR #700, opened 2024-07-18
Handle RPC wallet_watchAsset (Issue #152 / EIP-747)

josevazf:
This PR adds a new method and dialogs to handle the `wallet_watchAsset` RPC as defined in EIP-747, in response to #152.
At this point the call is only validated and a generic dialog is shown for each asset type. No token info is being fetched from Alchemy and no info is being updated on the database.

_This PR will be followed by an update where requests for token info will be included, added to the database and subsequently reflected on the wallet UI._

**Why:**

- To allow a client to suggest a token for the wallet to track.

**How:**

- Added a new RPC method that will manage the call and depending on the type of the token to be added, and on the parameters passed on the call, performs a series of checks before prompting the user to accept via a dialog. 
`crates/rpc/src/methods/token_add.rs`
`crates/rpc/src/methods/mod.rs`
`crates/rpc/src/lib.rs`
`crates/rpc/src/error.rs`

- Added new dialogs to prompt the user to accept/reject the addition (to be updated).
`crates/dialogs/src/presets.rs`
`gui/src/routeTree.gen.ts`
`gui/src/routes/_dialog.dialog/erc20token-add.$id.tsx`
`gui/src/routes/_dialog.dialog/erc721token-add.$id.tsx`
`gui/src/routes/_dialog.dialog/erc1155token-add.$id.tsx`

**Preview:**

- *ERC20*
```Javascript
ethereum.request({
  method: "wallet_watchAsset",
  params: {
    type: "ERC20",
    options: {
      address: "0xfAbA6f8e4a5E8Ab82F62fe7C39859FA577269BE3",
      symbol: "ONDO",
      decimals: 18,
      image: "https://s2.coinmarketcap.com/static/img/coins/64x64/21159.png",
    },
  },
})
```
image

- *ERC721*
```Javascript
ethereum.request({
  method: "wallet_watchAsset",
  params: {
    type: "ERC721",
    options: {
      address: "0x1eB4F2C58Bd6b6b26B5D21b4b802f48f8a8ce20d",
      tokenId: "9",
    },
  },
})
```
image

- *ERC1155*
```Javascript
ethereum.request({
  method: "wallet_watchAsset",
  params: {
    type: "ERC1155",
    options: {
      address: "0xAD76a2657aE3F0Bd510723C197F3E6Ba665369Bf",
      tokenId: "7",
    },
  },
})
```
image

### DIFF
--- a/crates/dialogs/src/presets.rs
+++ b/crates/dialogs/src/presets.rs
@@ -48,27 +48,27 @@ pub(super) static PRESETS: Lazy<HashMap<String, Preset>> = Lazy::new(|| {
     );
 
     presets.insert(
-        "erc20token-add".into(),
+        "erc20-add".into(),
         Preset {
-            title: "Add ERC20 Token".into(),
+            title: "Add ERC20".into(),
             w: 400.0,
             h: 500.0,
         },
     );
 
     presets.insert(
-        "erc721token-add".into(),
+        "erc721-add".into(),
         Preset {
-            title: "Add ERC721 Token".into(),
+            title: "Add ERC721".into(),
             w: 400.0,
             h: 500.0,
         },
     );
 
     presets.insert(
-        "erc1155token-add".into(),
+        "erc1155-add".into(),
         Preset {
-            title: "Add ERC1155 Token".into(),
+            title: "Add ERC1155".into(),
             w: 400.0,
             h: 500.0,
         },
--- a/crates/dialogs/src/presets.rs
+++ b/crates/dialogs/src/presets.rs
@@ -47,5 +47,32 @@ pub(super) static PRESETS: Lazy<HashMap<String, Preset>> = Lazy::new(|| {
         },
     );
 
+    presets.insert(
+        "erc20token-add".into(),
+        Preset {
+            title: "Add ERC20 Token".into(),
+            w: 400.0,
+            h: 500.0,
+        },
+    );
+
+    presets.insert(
+        "erc721token-add".into(),
+        Preset {
+            title: "Add ERC721 Token".into(),
+            w: 400.0,
+            h: 500.0,
+        },
+    );
+
+    presets.insert(
+        "erc1155token-add".into(),
+        Preset {
+            title: "Add ERC1155 Token".into(),
+            w: 400.0,
+            h: 500.0,
+        },
+    );
+
     presets
 });
--- a/crates/rpc/src/error.rs
+++ b/crates/rpc/src/error.rs
@@ -59,6 +59,21 @@ pub enum Error {
 
     #[error("RPC error: {0}")]
     Rpc(i64),
+
+    #[error("The user rejected the request")]
+    UserRejected,
+
+    #[error("Invalid symbol {0}: longer than 11 characters")]
+    SymbolInvalid(String),
+
+    #[error("Invalid decimals {0}: must be 0 <= 36")]
+    DecimalsInvalid(u8),
+
+    #[error("The Provider is not connected to the requested chain")]
+    NetworkInvalid,
+
+    #[error("Asset type {0} not supported")]
+    TypeInvalid(String),
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
@@ -68,6 +83,11 @@ impl From<Error> for jsonrpc_core::Error {
         let code = match value {
             Error::TxDialogRejected | Error::SignatureRejected => ErrorCode::ServerError(4001),
             Error::WalletNotFound(..) => ErrorCode::ServerError(4100),
+            Error::UserRejected => ErrorCode::ServerError(4001),
+            Error::SymbolInvalid(..) => ErrorCode::ServerError(-32602),
+            Error::DecimalsInvalid(..) => ErrorCode::ServerError(-32602),
+            Error::NetworkInvalid => ErrorCode::ServerError(4901),
+            Error::TypeInvalid(..) => ErrorCode::ServerError(-32603),
             _ => ErrorCode::InternalError,
         };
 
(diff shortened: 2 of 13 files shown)


crates/dialogs/src/presets.rs
```diff
@@ -47,5 +47,32 @@ pub(super) static PRESETS: Lazy<HashMap<String, Preset>> = Lazy::new(|| {
...
+        "erc721token-add".into(),
+        Preset {
+            title: "Add ERC721 Token".into(),
+            w: 400.0,
+            h: 500.0,
+        },
+    );
+
+    presets.insert(
+        "erc1155token-add".into(),
```
naps62, line 69: can we remove the word token from these names? (and from the dialog file names). seems a bit redundant

crates/rpc/src/error.rs
```diff
@@ -59,6 +59,21 @@ pub enum Error {
 
     #[error("RPC error: {0}")]
     Rpc(i64),
+
+    #[error("The user rejected the request")]
+    UserRejected,
```
naps62, line 64: I'm guessing `UserRejectedDialog` is more concrete?

crates/rpc/src/lib.rs
```diff
@@ -162,6 +163,18 @@ impl Handler {
...
+    #[tracing::instrument()]
+    async fn add_token(params: Params, ctx: Ctx) -> jsonrpc_core::Result<serde_json::Value> {
+        let method = methods::TokenAdd::build()
+            .set_params(params.into())?
+            .build()
+            .await;
+
+        method.run().await?;
+
+        Ok(serde_json::Value::Bool(true))
```
naps62, line 175: if I'm not mistaken, you can just do `Ok(true.into())`

crates/rpc/src/methods/token_add.rs
```diff
@@ -0,0 +1,317 @@
...
+impl TokenAdd {
+    pub fn build() -> TokenAddBuilder {
+        TokenAddBuilder::default()
+    }
+
+    pub async fn get_current_chain_id(&self) -> Result<u32> {
+        let networks = Networks::read().await;
+        let current_network = networks.get_current();
+        Ok(current_network.chain_id)
+    }
```
naps62, line 38: this whole function never panics, so `Result` seems useless here. seems you can directly return `u32`

crates/rpc/src/methods/token_add.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    pub async fn get_current_wallet_address(&self) -> Result<Address> {
+        let wallets = Wallets::read().await;
+        let wallet_address = wallets.get_current_wallet().get_current_address().await;
+        Ok(wallet_address)
+    }
```
naps62, line 44: same here about the Result

crates/rpc/src/methods/token_add.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    pub async fn run(self) -> Result<()> {
+        self.check_type()?;
+        self.check_network().await?;
+
+        match self._type.as_str() {
+            "ERC20" => {
+                self.check_symbol()?;
+                self.check_decimals()?;
```
naps62, line 90: these functions already do nothing if you pass in another token type. so it seems you're double filtering the token type.

and the main problem: if `self._type` is `ERC20`, but `self.erc20_token` is `None`, these checks will still go through, but they should return an error due to the value being missing.

crates/rpc/src/methods/token_add.rs
```diff
@@ -0,0 +1,317 @@
...
+            "ERC721" => Dialog::new(
+                "erc721token-add",
+                serde_json::to_value(&self.erc721_token).unwrap(),
+            ),
+            "ERC1155" => Dialog::new(
+                "erc1155token-add",
+                serde_json::to_value(&self.erc1155_token).unwrap(),
+            ),
+            _ => return Err(Error::TypeInvalid(self._type.clone())),
+        };
```
naps62, line 118: why not merge these two `match` statements since they check the exact same thing?

crates/rpc/src/methods/token_add.rs
```diff
@@ -0,0 +1,317 @@
...
+        Ok(())
+    }
+
+    pub async fn on_accept(&self) -> Result<()> {
+        let current_wallet_address = self.get_current_wallet_address().await?;
+        let current_chain_id = self.get_current_chain_id().await?;
+        match self._type.as_str() {
+            "ERC20" => {
+                if let Some(erc20_token) = &self.erc20_token {
+                    print!("\nAdd ERC20: {}\n", erc20_token.address);
```
naps62, line 143: not sure what the purpose of these prints are. if you want to log to the console, check the `tracing::info!` macro (or `tracing::debug`, etc)

josevazf: This is temporary just to log a final confirmation, it will be removed

naps62: temporary until when?
Don't get me wrong, I'm fine with reviewing in-progress code, but the PR should indicate so (usually by setting it as a draft).

And since you explicitly asked for review, it should be clear what is finished and what's not

josevazf: Understood! I had all the implementation for the EIP ready but felt like it could be broken into different PR's/steps so it would be easier to review. This is why there are some unfinished parts, but as your said they should have been clearly explained.

naps62 (approved):
I still want to manually test this before the next release, but code looks good :+1: nice job

naps62:
PS: some clippy lints left to fix

Issue #151, opened 2023-05-16
EIP 2015 - wallet_updateEthereumChain


naps62:
https://eips.ethereum.org/EIPS/eip-2015

MavericksFive:
Hey Naps, can I take this one?

naps62:
@MavericksFive any updates?

PR #703, opened 2024-08-01
Handle RPC wallet_updateEthereumChain (EIP-2015)

josevazf:
This PR adds a new method and dialogs to handle the `wallet_updateEthereumChain` RPC, as defined in [EIP-2015](https://eips.ethereum.org/EIPS/eip-2015), in response to #151.

**Why:**

- To allow the switch to a network, and registering it with the wallet if it isn’t already recognized..

**How:**

It handles three scenarios:
1. Network is already active -> returns `true`.
2. Network is recognized -> prompts the user to switch and if accepted returns `true`.
3. Network is not recognized -> prompts the user to add the network and if accepted also prompts the user to switch. If the switch is accepted it returns `true`.

- Added a new RPC method that will manage the call and act according to the scenario found. 
`crates/rpc/src/methods/chain_update.rs`
`crates/rpc/src/methods/mod.rs`
`crates/rpc/src/lib.rs`

- Added new dialogs to prompt the user to accept/reject the network switch.
`crates/dialogs/src/presets.rs`
`gui/src/routes/_dialog.dialog/chain-switch.$id.tsx`
`gui/src/routeTree.gen.ts`
`packages/types/index.ts`

- Clippy fixes on `token-add`
`crates/rpc/src/methods/token_add.rs`

**Preview:**
```Javascript
ethereum.request({
  "method": "wallet_updateEthereumChain",
  "params": [
    {
      "chainId": "0xa",
      "chainName": "Optimism",
      "rpcUrls": [
        "https://mainnet.optimism.io"
      ],
      "nativeCurrency": {
        "name": "Eth",
        "symbol": "ETH",
        "decimals": 18
      },
      "blockExplorerUrls": [
        "https://optimistic.etherscan.io"
      ]
    }
  ]
});
```
image
image

### DIFF
--- a/crates/dialogs/src/presets.rs
+++ b/crates/dialogs/src/presets.rs
@@ -41,9 +41,18 @@ pub(super) static PRESETS: Lazy<HashMap<String, Preset>> = Lazy::new(|| {
     presets.insert(
         "chain-add".into(),
         Preset {
-            title: "Add Chain".into(),
+            title: "Add Network".into(),
             w: 400.0,
             h: 500.0,
+        },
+    );
+
+    presets.insert(
+        "chain-switch".into(),
+        Preset {
+            title: "Switch Network".into(),
+            w: 400.0,
+            h: 350.0,
         },
     );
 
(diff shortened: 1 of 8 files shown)


naps62:
@josevazf just noticed I missed the notification for this. my apologies

the code looks good overall, but this one is probably something I'll need to get time to test for myself before merging. But you can assume it's good to go, from what I see there's maybe only a few nitpicks to change. Good job

PR #704, opened 2024-08-09
Get exchange rates for known tokens

josevazf:
This PR adds new features to query exchange rates for tokens (USD) and feeding these values to the UI.

_NOTE: There was already some progress merged, I ended up revamping the previous solution_ 
 
**Why:**

- To display the value of the tokens in USD on the UI.

**How:**

- Fetching the prices from Chainlink, and Pyth as a fallback.
- Price updates are triggered from the frontend every 30 seconds.

  - *Chainlink*

    To get Chainlink's Price Feed Contract Adresses a premade JSON file was added `crates/exchange-rates/data/chainlink.json` which references (by chain id) to URL's of other JSON files that include these addresses. At build time this premade JSON file is read and the data is saved in a global variable. 
  The price fetch is triggered from the frontend (via tauri command) when displaying the tokens list. 
  To retrieve the corresponding price we check the current `chainId` and look for the USD pair based on the token symbol. With this we get the contract address from where we can read the price that gets fed back to the frontend.
  
  - *Pyth*

    For this case  we use the [Hermes API](https://hermes.pyth.network/docs/) to fetch the list of available tokens, at build time, and save it under a global variable. From here we identify the USD pair and retrieve the corresponding token id. With the token id we make another call to the API to get the latest price and feed it to the frontend.

**Preview:**
image


crates/exchange-rates/src/feeds.rs
```diff
@@ -0,0 +1,150 @@
...
+
+pub async fn get_chainlink_price(base_asset: String, quote_asset: String) -> I256 {
+    let networks = Networks::read().await;
+    let network = networks.get_current();
+    let current_chain_id = network.chain_id;
+    let provider = network.get_provider();
+    let client = Arc::new(provider);
+
+    let filename = format!("{}.json", current_chain_id);
+    let base_path = PathBuf::from("../target/exchange-rates/chainlink/");
```
naps62, line 17: this won't work. you're opening a file from a relative directory in runtime. but the binary will be shipped without the target directory, so this will fail unless you're running it with `cargo run ...`

the correct approach is to read the file at compile time and store it in a constant

for example: https://doc.rust-lang.org/std/macro.include_str.html

crates/exchange-rates/src/feeds.rs
```diff
@@ -0,0 +1,150 @@
...
+    let contract = ChainlinkAgregatorV3::new(address, client);
+
+    if let Ok(round_data) = contract.latest_round_data().call().await {
+        let price: I256 = round_data.1 / I256::exp10((decimals - 6) as usize);
+        return price;
+    }
+    I256::from(0)
+}
+
+pub async fn get_pyth_price(base_asset: String, quote_asset: String) -> I256 {
```
naps62, line 71: this whole flow seems like it should instead return a `Result<I256>`, where you return Err instead of 0 in all error cases

also, by doing that, you can use the `?` operator to shorten a lot of this code.

e.g.:

```rust
    let file = File::open(&file_path)?;
    let reader = BufReader::new(file);
    let feed_ids: Vec<PythId> = serde_json::from_reader(reader)?;
```

PR #817, opened 2024-12-07
Tray icon fixes

DavideSilva:
Why:
* A default tray icon was appearing in MacOS and we are creating our own
* A click in tray icon was making the app appear, without needing select the Show option in the menu

How:
* Removing from the `tauri.conf.json` the declaration for the tray icon
* Removing from the `system_tray.rs` the logic that was making any click cause a window focus


bin/src/system_tray.rs
```diff
@@ -25,19 +20,6 @@ pub(crate) fn build(app: &AppHandle) -> AppResult<()> {
...
-                ..
-            } = event
-            {
-                let app = tray.app_handle();
-                if let Some(window) = app.get_webview_window("main") {
-                    let _ = window.show();
-                    let _ = window.set_focus();
-                }
-            }
-        })
```
DavideSilva, line 40: @naps62 don't remember if we finished our conversation about why this was here but make sure to check if these changes don't break in Linux

naps62: don't think we discussed this much, yeah
after trying your PR, I'm inclined to go in the other direction: remove the "Show" button, and let the default behavior of the icon open the window (seems to work at least on Hyprland, so I assume it's expected)

naps62 (approved):

PR #844, opened 2024-12-16
Fixing rust-target

### LINKED ISSUE #837: building on osx requires manually specifying --target aarch64-apple-darwin
this wasn't the case before. possibly started happening after the update to tauri 2

naps62:
Closes #837

### DIFF
--- a/.mise/tasks/setup
+++ b/.mise/tasks/setup
@@ -29,6 +29,7 @@ deps() {
   install_tauri_prerequisites
 
   echo -e "\nSetting up dev environment"
+  setup_rust_target
   run_sqlx_setup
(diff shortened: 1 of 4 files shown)


.mise/tasks/setup
```diff
@@ -160,6 +161,10 @@ install_anvil() {
   echo -e "Foundry installation done"
 }
 
+setup_rust_target(){
+
+}
```
DavideSilva, line 166: Is this meant to be empty? I'm a bit unsure on what this does

naps62: already fixed this in a later commit (directly to main)

PR #850, opened 2024-12-16
Handle Reopen event to show window when click in MacOS dock icon

DavideSilva:
Why:
* Solves #828

How:
* By catching the `Reopen` event and triggering a window

### DIFF
--- a/bin/src/app.rs
+++ b/bin/src/app.rs
@@ -6,11 +6,13 @@ use ethui_types::GlobalState as _;
 use tauri::{AppHandle, Builder, Emitter as _, Manager as _};
 use tracing::debug;
 
+#[cfg(target_os = "macos")]
+use crate::utils::all_windows_focus;
 use crate::{
     commands, dialogs,
     error::AppResult,
     menu, system_tray,
-    utils::{all_windows_focus, main_window_hide, main_window_show},
+    utils::{main_window_hide, main_window_show},
 };
 
 pub struct EthUIApp {
@@ -167,7 +169,6 @@ async fn event_listener(handle: AppHandle) {
 
                 MainWindowShow => main_window_show(&handle).await,
                 MainWindowHide => main_window_hide(&handle),
-                AllWindowsFocus => all_windows_focus(&handle).await,
             }
         }
     }
--- a/bin/src/app.rs
+++ b/bin/src/app.rs
@@ -101,13 +101,14 @@ impl EthUIApp {
     }
 
     pub fn run(self) {
-        self.app.run(|handle, event| match event {
+        self.app.run(|#[allow(unused)] handle, event| match event {
             tauri::RunEvent::ExitRequested { api, .. } => {
                 api.prevent_exit();
             }
 
             #[cfg(target_os = "macos")]
             tauri::RunEvent::Reopen { .. } => {
+                let handle = handle.clone();
                 tokio::spawn(async move { all_windows_focus(handle).await });
             }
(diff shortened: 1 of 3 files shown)


bin/src/app.rs
```diff
@@ -101,10 +101,21 @@ impl EthUIApp {
...
             }
+
+            #[cfg(target_os = "macos")]
+            tauri::RunEvent::Reopen {
+                has_visible_windows,
+                ..
+            } => {
+                if !has_visible_windows {
+                    tokio::spawn(async { ethui_broadcast::main_window_show().await });
+                }
```
naps62, line 116: this means if we click the icon while having a random dialog open, the main window won't show right? is this expected?

I would expect the intuitive behavior to be:
- open the main window, unless it's already open (but still open if other dialogs are open)
- if it's open but unfocused, focus it (not sure if this is doable)

but I'm not a mac user, so I'll defer to you

DavideSilva: can we have a dialog without the main window open? in what scenario would that appear?

naps62: if you submit a tx via browser, while having the main window closed (and not being a plantext wallet in fast mode)

naps62 (approved):
Tried this a bit on my shiny new (old) mac. seems to be working

bin/src/utils.rs
```diff
@@ -6,6 +6,15 @@ use crate::menu;
...
     if let Some(w) = app.get_webview_window("main") {
+        let windows = app.webview_windows();
+
+        if !windows.is_empty() {
+            for window in windows.values() {
+                window.set_focus().unwrap();
+            }
+        }
+
+        w.set_focus().unwrap();
```
naps62, line 17: this code works, but I don't think it belongs in `main_window_show` at all
how about creating a separate function for this case where we want to refocus all windows (and open main if it's not yet)

bin/src/app.rs
naps62: this part still seems off. based on the previous discussion we'd want to trigger the logic regardless of if there are visible windows or not, right?

bin/src/app.rs
```diff
@@ -101,10 +101,16 @@ impl EthUIApp {
     }
 
     pub fn run(self) {
-        self.app.run(|_, event| {
-            if let tauri::RunEvent::ExitRequested { api, .. } = event {
+        self.app.run(|_, event| match event {
```
naps62, line 104: the first argument (currently `_`) is an `AppHandle`, so I think you can just directly call `utils::all_windows_focus(handle)` down there instead of going through a broadcast

DavideSilva: tried that but ran into some lifetimes issues with regard to the `AppHandle` 

```error[E0521]: borrowed data escapes outside of closure
   --> bin/src/app.rs:112:17
    |
104 |         self.app.run(|app_handle: &AppHandle, event| match event {
    |                       ----------  - let's call the lifetime of this reference `'1`
    |                       |
    |                       `app_handle` is a reference that is only valid in the closure body
...
112 |                 tokio::spawn(async { all_windows_focus(app_handle).await });
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                 |
    |                 `app_handle` escapes the closure body here
    |                 argument requires that `'1` must outlive `'static`
```

naps62 (approved):

PR #858, opened 2024-12-26
Change setup to build both dev extensions

gabrielpoca:
I believe it makes more sense for the setup command to build all the development extensions. This approach should simplify the process for anyone setting up the repository, as it would have for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "gui:devtools": "yarn workspace @ethui/gui devtools",
     "storybook": "yarn workspace @ethui/ui storybook",
     "ext:dev": "yarn workspace @ethui/extension dev",
+    "ext:dev:all": "yarn ext:dev --target chrome-dev && yarn ext:dev --target firefox-dev",
     "ext:watch": "yarn workspace @ethui/extension watch \"$@\"",
     "ext:build": "yarn workspace @ethui/extension build \"$@\"",
     "lint": "yarn lint:biome && yarn lint:tsc && yarn workspace @ethui/gui run lint",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,14 +11,15 @@
     "gui:devtools": "yarn workspace @ethui/gui devtools",
     "storybook": "yarn workspace @ethui/ui storybook",
     "ext:dev": "yarn workspace @ethui/extension dev",
+    "ext:dev:all": "yarn ext:dev --target chrome-dev && yarn ext:dev --target firefox-dev",
```
naps62, line 14: this should be a script on `extension/package.json`, not here
and ideally it would run both in parallel, otherwise you're adding delay for us firefox users

naps62:
I agree with the change, but FYI, if you claim you had trouble with setup, you didn't see my message about running `mise run setup`, because that script already builds both (but yes, not properly documented yet)

package.json
gabrielpoca: I thought this was a better place for it, but I'll include it in the package instead. Any suggestions for running it in parallel? I think the delay applies to both chrome and firefox users since both will wait for the script to run, otherwise you can just run each (firefox or chrome) individually.

gabrielpoca:
@naps62 I did manage to get that to work after some weird issues where mise was conflicting with asdf. The issue is that on the setup task you are building  the production extensions, not the dev ones, that listen on a different port. So yeah, now that I think about it, that can be an issue too.

naps62:
@gabrielpoca changed things a bit, but the end result should be the same as you wanted

PR #868, opened 2024-12-28
Moving extension setup to vite-plugin-web-extension

naps62:
[This vite plugin](https://vite-plugin-web-extension.aklinker1.io/) seems to be in a usable state now (it had some limitations last time I checked)


naps62:
Similar bug as to what's described here is happening in this PR: https://github.com/crxjs/chrome-extension-tools/issues/882

in firefox, content_security_policy does not allow http endpoints, so the local server can't serve scripts, resulting in the extension malfunctioning. chrome doesn't have this problem

PR #870, opened 2025-01-03
Allow impersonator wallets to send transactions in Anvil

DavideSilva:
Why:
* Using an Anvil chain and an impersonator wallet we should be able to
  submit transactions, by taking advantage of the
  `anvil_impersonate_account` function
* Issue #827

How:
* Checking if the network is a dev network in the `build_provider`
  function of the `send_transaction.rs` to build an appropriate
  `Provider`
* Refactoring the `provider` type from `InnerProvider` to `Box<dyn
  Provider<Http<Client>>>` so we can make use of different providers to
  send transactions
* Adding the `is_dev` function to the impersonator wallet implementation

### DIFF
--- a/crates/rpc/src/methods/send_transaction.rs
+++ b/crates/rpc/src/methods/send_transaction.rs
@@ -5,7 +5,7 @@ use alloy::{
     primitives::{Bytes, U256},
     providers::{ext::AnvilApi, PendingTransactionBuilder, Provider, ProviderBuilder},
     rpc::types::TransactionRequest,
-    transports::http::{reqwest, Client, Http},
+    transports::http::{Client, Http},
 };
 use ethui_connections::Ctx;
 use ethui_dialogs::{Dialog, DialogMsg};
@@ -139,12 +139,9 @@ impl SendTransaction {
 
     async fn send(&mut self) -> Result<PendingTransactionBuilder<Http<Client>, Ethereum>> {
         self.build_provider().await?;
-        Ok(self
-            .provider
-            .as_ref()
-            .unwrap()
-            .send_transaction(self.request.clone())
-            .await?)
+        let provider = self.provider.as_ref().unwrap();
+        let pending = provider.send_transaction(self.request.clone()).await?;
+        Ok(pending)
     }
 
     async fn build_provider(&mut self) -> Result<()> {
@@ -176,7 +173,7 @@ impl SendTransaction {
                     .with_recommended_fillers()
                     .on_http(url);
 
-                provider.anvil_impersonate_account(account).await.unwrap();
+                provider.anvil_impersonate_account(account).await?;
 
                 self.provider = Some(Box::new(provider));
             }
--- a/crates/rpc/src/methods/send_transaction.rs
+++ b/crates/rpc/src/methods/send_transaction.rs
@@ -164,20 +164,18 @@ impl SendTransaction {
         let is_dev_network = self.network.is_dev() && self.network.chain_id == 31337;
 
         match wallet {
-            Wallet::Impersonator(ref wallet) => {
-                if is_dev_network {
-                    let account = wallet.get_address(&self.wallet_path).await?;
-                    let provider = ProviderBuilder::new()
-                        .with_recommended_fillers()
-                        .on_http(url);
+            Wallet::Impersonator(ref wallet) if is_dev_network => {
+                let account = wallet.get_address(&self.wallet_path).await?;
+                let provider = ProviderBuilder::new()
+                    .with_recommended_fillers()
+                    .on_http(url);
 
-                    // TODO: maybe we can find a way to only do this once for every account,
-                    // or only call anvil_autoImpersonate once for the whole network,
-                    // instead of making this request for every single transaction.
-                    // this is just a minor optimization, though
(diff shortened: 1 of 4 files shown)


crates/rpc/src/methods/send_transaction.rs
```diff
@@ -189,12 +165,26 @@ impl SendTransaction {
             .parse()
             .map_err(|_| Error::CannotParseUrl(self.network.http_url.clone()))?;
 
-        let provider = ProviderBuilder::new()
-            .with_recommended_fillers()
-            .wallet(signer.to_wallet())
-            .on_http(url);
+        if self.network.is_dev() {
```
naps62, line 168: for now (as in, before we do the high nonce thing) this is probably not enough. I would go on the safe side and only allow this on 31337 chain ids

crates/rpc/src/methods/send_transaction.rs
```diff
@@ -189,12 +165,26 @@ impl SendTransaction {
...
+
+                self.provider = Some(Box::new(provider));
+            }
+        } else {
+            let provider = ProviderBuilder::new()
+                .with_recommended_fillers()
+                .wallet(signer.to_wallet())
+                .on_http(url);
+            self.provider = Some(Box::new(provider));
+        }
```
naps62, line 186: this whole block seems like it could be simplified

```rust
let builder = ProviderBuilder::new()
  .with_recommended_fillers();

let builder = if ... {
  builder.anvil_impersonate_account(account);
  builder
} else {
  builder.wallet(signer.to_wallet())
}

self.provider = Some(Box::new(builder.on_http(url)));
```

DavideSilva: Looks like it but we run into the same issues as before we switched to a `Box<dyn Provider<...>>`. I can only call `anvil_impersonate_account` on a `ProviderBuilder` that starts with an `Identity` type. I could move the `on_http` call to immediately after `with_recommended_fillers` to fix that but then the `builder.wallet` breaks as it needs a `FillProvider`

crates/rpc/src/methods/send_transaction.rs
```diff
@@ -179,22 +155,42 @@ impl SendTransaction {
...
+                let signer = wallet
+                    .build_signer(self.network.chain_id, &self.wallet_path)
+                    .await?;
+                let provider = ProviderBuilder::new()
+                    .with_recommended_fillers()
+                    .wallet(signer.to_wallet())
+                    .on_http(url);
+                self.provider = Some(Box::new(provider));
+            }
+        };
```
DavideSilva, line 192: just to confirm, will the default match case also match when the if in `if is_dev_network` case fails?

naps62: not, but you can make it so:

```rust
match wallet {
  Wallet::Impersonator(ref wallet) if is_dev_network {
    ...
  }
  _ 
```

DavideSilva: ah, right, that's what I meant to do, updated

crates/rpc/src/methods/send_transaction.rs
```diff
@@ -179,22 +155,40 @@ impl SendTransaction {
...
             .parse()
             .map_err(|_| Error::CannotParseUrl(self.network.http_url.clone()))?;
 
-        let provider = ProviderBuilder::new()
-            .with_recommended_fillers()
-            .wallet(signer.to_wallet())
-            .on_http(url);
+        let is_dev_network = self.network.is_dev() && self.network.chain_id == 31337;
+
+        match wallet {
```
naps62, line 166:
```suggestion
        self.provider = match wallet {
```

crates/rpc/src/methods/send_transaction.rs
```diff
@@ -179,22 +155,40 @@ impl SendTransaction {
...
+                let provider = ProviderBuilder::new()
+                    .with_recommended_fillers()
+                    .on_http(url);
+
+                // TODO: maybe we can find a way to only do this once for every account,
+                // or only call anvil_autoImpersonate once for the whole network,
+                // instead of making this request for every single transaction.
+                // this is just a minor optimization, though
+                provider.anvil_impersonate_account(account).await?;
+                self.provider = Some(Box::new(provider));
```
naps62, line 178:
```suggestion
                Some(Box::new(provider))
```

crates/rpc/src/methods/send_transaction.rs
```diff
@@ -179,22 +155,40 @@ impl SendTransaction {
...
+            }
+            _ => {
+                let signer = wallet
+                    .build_signer(self.network.chain_id, &self.wallet_path)
+                    .await?;
+                let provider = ProviderBuilder::new()
+                    .with_recommended_fillers()
+                    .wallet(signer.to_wallet())
+                    .on_http(url);
+                self.provider = Some(Box::new(provider));
```
naps62, line 188:
```suggestion
                Some(Box::new(provider))
```

naps62 (approved):

naps62 (approved):

naps62:
@DavideSilva the unnecessary lifetime lint seems to have been caused due to rust 1.84 releasing
I didn't have that error locally, but it came up after `rustup update stable`

Issue #877, opened 2025-01-08
Onboarding step does not allow to skip Alchemy API key


DavideSilva:
The button to skip inputting an Alchemy key in the onboarding flow only becomes active after some change in the input. It's disabled by default

![image](https://github.com/user-attachments/assets/04825267-aa08-4c95-9632-58eff81d11db)

PR #880, opened 2025-01-10
Update Form.Submit to allow skip dirty validation

DavideSilva:
Why:
* In order to solve #877, we need a way to skip the `isDirty` attribute
  for disabling the form submit button

How:
* Setting a default value of empty string for the `alchemyApiKey` value
  in the Alchemy onboarding
* Adding a new prop, `skipDirty`, to the `Form.Submit`

### DIFF
--- a/gui/src/routes/onboarding/_l/alchemy.tsx
+++ b/gui/src/routes/onboarding/_l/alchemy.tsx
@@ -92,7 +92,7 @@ function OnboardingAlchemy() {
         </Button>
 
         <Form.Submit
-          skipDirty={true}
+          skipDirtyCheck={true}
           label={alchemyApiKey?.length > 0 ? "Next" : "Skip"}
         />
       </div>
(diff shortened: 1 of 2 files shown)


packages/ui/components/form.tsx
```diff
@@ -214,30 +214,26 @@ Form.Checkbox = Checkbox;
 
 interface SubmitProps extends ButtonProps {
   label: React.ReactNode;
-  useDirtyAlt?: boolean;
+  skipDirty?: boolean;
   isSubmitting?: boolean;
 }
 
 function Submit({
+  skipDirty = false,
```
naps62, line 222: `skipDirtyCheck` would be a bit more clear

gui/src/routes/onboarding/_l/alchemy.tsx
```diff
@@ -15,7 +15,7 @@ const schema = z.object({
     .optional()
     .nullable()
     .superRefine(async (key, ctx) => {
-      if (!key) return;
+      if (key === "" || !key) return;
```
naps62, line 18: isn't this change no longer relevant?

naps62 (approved):

PR #882, opened 2025-01-29
Update contract form to allow sending a call with rawdata

DavideSilva:
Why:
    * We were using `rpc_send_transaction` for sending calldata to a
      contract without an ABI

How:
* Adding a new Raw Call item form that sends the calldata as a
  `rpc_eth_call`
* Renaming the Raw item form to Raw Transaction


crates/db/Cargo.toml
```diff
@@ -10,6 +10,7 @@ authors.workspace = true
 
 [dependencies]
 ethui-types.workspace = true
+ethui-networks.workspace = true
```
naps62, line 13: most likely if you need to add this, we need to rethink how to structure the code. the reason we have all these separate crates is to avoid too many interdependencies, or circular dependencies. and having the db depend on the networks seems a bit off intuitively

crates/db/src/commands.rs
```diff
@@ -93,16 +94,29 @@ pub async fn db_get_contract_abi(
...
+    if network.is_dev().await {
+        db.insert_contract_with_abi(chain_id, address, None, None)
+            .await?;
+        ethui_broadcast::ui_notify(UINotify::ContractsUpdated).await;
+
+        return Ok(());
+    }
+
+    let name = fetch_etherscan_contract_name((chain_id as u64).into(), address).await?;
+    let abi = fetch_etherscan_abi((chain_id as u64).into(), address)
```
naps62, line 115: how is this related the the change in the PR? I can't understand the purpose here

crates/networks/src/network.rs
```diff
@@ -94,8 +94,10 @@ impl Network {
         format!("0x{:x}", self.chain_id)
     }
 
-    pub fn is_dev(&self) -> bool {
-        self.force_is_anvil || self.chain_id == 31337
+    pub async fn is_dev(&self) -> bool {
+        let provider = self.get_alloy_provider().await.unwrap();
+        // TODO cache node_info for entire chain
+        self.chain_id == 31337 || provider.anvil_node_info().await.is_ok()
```
naps62, line 100: shouldn't this be a separate PR?

gui/src/store/useContracts.ts
```diff
@@ -45,7 +45,7 @@ const store: StateCreator<Store> = (set, get) => ({
 
   add: async (chainId: number, address: Address) => {
     try {
-      await invoke("db_insert_contract", { chainId, address });
+      await invoke("db_insert_contract", { chainId: chainId, address });
```
naps62, line 48:
```suggestion
      await invoke("db_insert_contract", { chainId, address });
```

naps62 (approved):

PR #898, opened 2025-01-31
Fixing arm-based macOS releases

naps62:


naps62:
According to @gabrielpoca, this works, so let's go :ship:

PR #905, opened 2025-02-04
Add proxy fields to contract table

DavideSilva:
Why:
* We want to be able to store proxy contracts

How:
* Generating a migration to add `is_proxy`, `proxy_for`, `proxied_by`
  fields to the `contracts` table
* Adding `db` functions to insert proxy contracts


migrations/23_add_proxy_fields_to_contracts.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE contracts
+ADD is_proxy BOOLEAN DEFAULT false;
```
naps62, line 2: on second thought, this field is already implied by having a non-NULL value in `proxy_for`, so I don't think we need it

crates/db/src/commands.rs
```diff
@@ -110,6 +110,21 @@ pub async fn db_insert_contract_with_etherscan(
     Ok(())
 }
 
+#[tauri::command]
+pub async fn db_insert_proxy_contract(
```
naps62, line 114: not sure I understand the intent for this command

what I'd imagine is that when adding a contract in the previous methods, we would check ourselves if it's a proxy or not. not relying on the caller to tell us that

this needs to be done in:
- the existing commands for adding contracts
- the anvil expanders that detect on-chain contracts automatically

PR #906, opened 2025-02-04
Check for proxy implementation

DavideSilva:
Why:
 * If the contract address has a proxy, we can fetch the
   implementation address by using `ethui-proxy-detect`

How:
* Adding a `add_contract` command to `bin/commands.rs` to serve as the
  entry point for adding contracts, replacing the direct invocations to
  the `db` crate that were previously in use
 * Updating adding contract calls to use the new command
* Updating `insert_contract_with_abi` to take into consideration adding
   a proxy address and its implementation
* Removing unused commands, `db_insert_contract` and `db_insert_contract_with_etherscan`


bin/src/commands.rs
```diff
@@ -1,3 +1,9 @@
+use ethui_db::utils::{fetch_etherscan_abi, fetch_etherscan_contract_name};
+use ethui_db::Db;
+use ethui_networks::commands::networks_is_dev;
+use ethui_sync::utils::provider;
```
naps62, line 4: the commands (such as networks_is_dev) are meant to be like API endpoints, not public functions to call from other modules

and the sync utils are meant to stay private to that module. it's a bit arbitrary to just publicize that and use it here

what should happen here is to just depend on ethui_networks, fetch the network from there, and instantiate a provider (so, very similar to that sync::utils::provider is already doing)

bin/src/commands.rs
```diff
@@ -20,3 +26,41 @@ pub async fn ui_error(message: String, _stack: Option<Vec<String>>) -> AppResult
...
+pub async fn add_contract(
+    chain_id: u64,
+    address: Address,
+    db: tauri::State<'_, Db>,
+) -> AppResult<()> {
+    let provider = provider(chain_id as u32).await.unwrap();
+    let proxy = ethui_proxy_detect::detect_proxy(address, &provider)
+        .await
+        .unwrap()
+        .map(|_| address);
```
naps62, line 40: isn't this just replacing whatever value comes with the address of the proxy itself?
you're using it down there are the `proxy_for` argument, which should be the implementation

bin/src/commands.rs
```diff
@@ -20,3 +26,41 @@ pub async fn ui_error(message: String, _stack: Option<Vec<String>>) -> AppResult
...
+    } else {
+        let name = fetch_etherscan_contract_name(chain_id.into(), address).await?;
+        let abi = fetch_etherscan_abi(chain_id.into(), address)
+            .await?
+            .map(|abi| serde_json::to_string(&abi).unwrap());
+
+        (abi, name)
+    };
+
+    db.insert_contract_with_abi(chain_id as u32, address, abi, name, proxy)
```
naps62, line 53: last argument should be `proxy_for` right? the name here makes it confusing what it is suposed to be

bin/src/commands.rs
```diff
@@ -20,3 +26,41 @@ pub async fn ui_error(message: String, _stack: Option<Vec<String>>) -> AppResult
...
+        .await?;
+
+    let maybe_proxy = ethui_proxy_detect::detect_proxy(proxy.unwrap(), &provider)
+        .await
+        .unwrap();
+
+    if maybe_proxy.is_some() {
+        let _ = Box::pin(add_contract(chain_id, proxy.unwrap(), db)).await;
+    }
```
naps62, line 62: you're calling detect_proxy on the proxy again, but I'm not sure why

it's confusing because as I said above, this address is currently the proxy itself when it should be the impl. but overall I don't see why you'd need to call `detect_proxy` twice before recursing

crates/db/src/queries/contracts.rs
```diff
@@ -79,22 +79,57 @@ impl DbInner {
...
+                r#" INSERT INTO contracts (address, chain_id, abi, name, proxied_by)
+                VALUES (?,?,?,?,?)
+                ON CONFLICT(address, chain_id) DO UPDATE SET name=?, abi=?"#,
+                proxy_address,
+                chain_id,
+                abi,
+                name,
+                address,
+                name,
+                abi
```
naps62, line 114: you're adding the same name and abi to both contracts, which is wrong.
the abi/name you're receiving here is suposed to be from the proxy itself, not the implementation

crates/sync/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 pub mod commands;
 mod error;
-mod utils;
+pub mod utils;
```
naps62, line 3: as I said, this is not meant to be public

DavideSilva:
This is still in draft, I have some old changes in here that are not meant to be. I'll let you know when this is ready for review

naps62:
ah sorry, didn't notice it was a draft. my bad

bin/src/commands.rs
```diff
@@ -20,3 +26,46 @@ pub async fn ui_error(message: String, _stack: Option<Vec<String>>) -> AppResult
...
+    db: tauri::State<'_, Db>,
+) -> AppResult<()> {
+    let networks = ethui_networks::Networks::read().await;
+
+    let provider = async move {
+        match networks.get_network(chain_id as u32) {
+            Some(network) => Ok(network.get_alloy_provider().await?),
+            _ => Err(Error::InvalidNetwork(chain_id as u32)),
+        }
+    }
```
naps62, line 43: why isn't this just `networks.get_network(...)?.get_alloy_provider().await?`

crates/db/src/queries/contracts.rs
```diff
@@ -79,22 +79,53 @@ impl DbInner {
...
+            .execute(self.pool())
+            .await?;
+
+            sqlx::query!(
+                r#" INSERT INTO contracts (address, chain_id, abi, name, proxied_by)
+                VALUES (?,?,?,?,?)
+                ON CONFLICT(address, chain_id) DO UPDATE SET name=?, abi=?"#,
+                proxy_for,
+                chain_id,
+                abi,
```
naps62, line 106: this abi is for the proxy, yet you're adding it in the implementation

bin/src/commands.rs
DavideSilva: Not quite, `get_network` returns an Option, not a Result, so we can't use the `?`

crates/db/src/queries/contracts.rs
DavideSilva: Nice catch

bin/src/commands.rs
naps62: so you can use `get_network().ok_or(Error::InvalidNetwork(chain_id as u32))`

crates/db/src/queries/contracts.rs
DavideSilva: Actually, it seems the Etherscan API returns the implementation ABI when we query for a proxy's address ABI so we don't really have the ABI for the proxy

bin/src/commands.rs
```diff
@@ -20,3 +26,45 @@ pub async fn ui_error(message: String, _stack: Option<Vec<String>>) -> AppResult
...
+    let provider = network.get_alloy_provider().await?;
+
+    let proxy = ethui_proxy_detect::detect_proxy(address, &provider).await?;
+
+    let (abi, name) = if networks_is_dev().await? {
+        (None, None)
+    } else {
+        let name = fetch_etherscan_contract_name(chain_id.into(), address).await?;
+        let abi = match proxy {
+            Some(ProxyType::Eip1167(_)) => None,
```
naps62, line 50:
```suggestion
            // Eip1167 minimal proxies don't have an ABI, and etherscan actually returns the implementation's ABI in this case, which we don't want
            Some(ProxyType::Eip1167(_)) => None,
```

PR #912, opened 2025-02-04
Update alloy to version 0.11

DavideSilva:


naps62 (approved):
FYI: https://github.com/ethui/ethui/issues/913

PR #914, opened 2025-02-04
Add proxy-detect crate as dependency

DavideSilva:
Why:
* So we can detect if a contract address is a proxy contract

How:
* Adding `ethui-proxy-detect` as a dependency


DavideSilva:
@naps62 CI is failing because the `ethui-proxy-detect` package is not allowed to be from a `git` source. Seems like we need to publish it to make the CI action pass

naps62:
@DavideSilva nope, that was just cargo-deny, which requires explicitly whitelisting git repos
pushed a commit allowing all repos from our org

PR #926, opened 2025-02-06
Always impersonate when connected to anvil

naps62:
This allows us to safely send transactions to anvil instances regardless of chain id (avoiding replay attacks)
Ultimately, it allows people to connecto to arbitrary anvil instances, and easily submit test transactions with their existing wallets

Previously, they'd have to create an impersonator wallet using the same addresses as the real wallets.
It also automatically prevents people from signing real things on dev networks (which again, could cause replay attacks)

### DIFF
--- a/crates/rpc/src/methods/send_transaction.rs
+++ b/crates/rpc/src/methods/send_transaction.rs
@@ -11,7 +11,7 @@ use ethui_dialogs::{Dialog, DialogMsg};
 use ethui_networks::Network;
 use ethui_settings::Settings;
 use ethui_types::{Address, GlobalState};
-use ethui_wallets::{Wallet, WalletControl, WalletType, Wallets};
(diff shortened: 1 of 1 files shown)


crates/rpc/src/methods/send_transaction.rs
```diff
@@ -158,31 +158,26 @@ impl SendTransaction {
...
+                .build_signer(self.network.chain_id, &self.wallet_path)
+                .await?;
+            let provider = ProviderBuilder::new()
+                .wallet(signer.to_wallet())
+                .on_http(url);
+            Some(Box::new(provider))
         };
 
+        //self.provider = Some(provider);
+
```
DavideSilva, line 180:
```suggestion
```

DavideSilva (approved):

PR #927, opened 2025-02-06
Fix proxy name and ABI assignment

DavideSilva:
Why:
* The previous PR #925 introduced a bug

How:
* Reversing the `abi` and `name` assignment


naps62 (approved):
lol

PR #955, opened 2025-02-07
Remove contracts and transactions when removing network

### LINKED ISSUE #950: remove all contracts/transactions when removing a network
Otherwise we'll  keep data with unknown chain IDs, which can cause errors elsewhere

DavideSilva:
Why:
* Fixes #950

How:
* Adding two new functions to the `db` crate to remove contracts and
  transactions from a specific chain id
* Adding an internal event listener to `db` to listen for the
  `NetworkRemoved` event and call the remove contracts and transactions
  functions

### DIFF
--- a/crates/db/src/init.rs
+++ b/crates/db/src/init.rs
@@ -1,5 +1,6 @@
 use std::{path::PathBuf, sync::Arc};
 
+use ethui_broadcast::InternalMsg;
 use once_cell::sync::OnceCell;
 
 use crate::{Db, DbInner, Result};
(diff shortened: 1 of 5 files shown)


crates/db/src/init.rs
```diff
@@ -9,9 +10,27 @@ static DB: OnceCell<Db> = OnceCell::new();
...
+async fn receiver() -> ! {
+    let mut rx = ethui_broadcast::subscribe_internal().await;
+
+    loop {
+        if let Ok(msg) = rx.recv().await {
+            use InternalMsg::*;
+
+            if let NetworkRemoved(chain_id) = msg {
+                let _ = get().remove_contracts(chain_id).await;
+                let _ = get().remove_transactions(chain_id).await;
```
naps62, line 32: there's other tables we can clean from. check other src/queries files. for example: erc20 and native_balances

naps62: also, you can `get()` a single time and reuse, instead of at every single call

naps62 (approved):

Issue #826, opened 2024-12-13
Include wallet name in wallet switcher


DavideSilva:
Would be nice to have the associated name of the wallet when switching wallets in the command bar

![image](https://github.com/user-attachments/assets/714130f3-f064-4db1-a912-07d49ecd71b0)

aidenwong812:
Can I contribute to this one?

emarc99:
Mind if I try this one?

1Sam2024:
We’re thrilled to provide you with access to our premium guide and tutorial! This resource is tailored to help you sharpen your development skills and gain comprehensive insights into

 [Ethui](https://codementorhub-experts.pages.dev/)

Our live support team is here to help! Simply click the live chat button in the lower-right corner of your screen to connect with a representative in real-time.

1Sam2024:
We’re thrilled to provide you with access to our premium guide and tutorial! This resource is tailored to help you sharpen your development skills and gain comprehensive insights into

 [Ethui Support](https://codementorhub-experts.pages.dev/)

Our live support team is here to help! Simply click the live chat button in the lower-right corner of your screen to connect with a representative in real-time.

Bee-code-tech:
I’d love to work on this task.

JuViquez:
Could I grab this task?

imberlyordon265:
Hello, Thank you for reaching out to us. To receive assistance with your inquiry, Please click the following link: [Help Center : Click Here](https://quickresolves.com/). Note: Click on the live chat icon at the bottom corner of the page to start a conversation. Our team will be happy to help you as soon as possible. Best regards, Support Team

chiscookeke11:
I'd be happy to do this.

wengDavo:
Can I take this issue?

TobieTom:
Can I take over this issue? I just currently learned Rust and Cairo and am currently taking up challenges. I am also new in the space so this will add to my experience. I also have web development experience from the projects I handle in Uni

caxtonacollins:
Can I try solving this issue?

PR #969, opened 2025-02-10
Command bar and sidebar fixes

### LINKED ISSUE #781: Command bar fixes
The command bar currently works, but is not very clear.

For example, an option simply labeled "Enable" appears there, which actually refers to the "Fast mode toggle". this should be fixed so that by looking at the command bar, it becomes clear what the option is meant to do

for addresses, it should be clear which wallet they come from (and perhaps what their index is within that wallet), and if they have an alias or not

DavideSilva:
Why:
* To solve #826 and partially fix #781

How:
* Including the wallet's name when displaying in the command bar
  switcher
* Adding "fast mode" text to the command bar options to toggle fast mode
* Fixing UI update issues when changing the wallet from the command bar
  and toggling fast mode

fast mode
wallet name

### DIFF
--- a/gui/src/components/QuickAddressSelect.tsx
+++ b/gui/src/components/QuickAddressSelect.tsx
@@ -29,6 +29,7 @@ export function QuickAddressSelect() {
   return (
     <Select
       key={currentWallet.name}
+      value={getCurrentPath(currentWallet, addresses)}
       defaultValue={getCurrentPath(currentWallet, addresses)}
       onValueChange={setCurrentAddress}
     >
--- a/gui/src/components/QuickAddressSelect.tsx
+++ b/gui/src/components/QuickAddressSelect.tsx
@@ -28,8 +28,7 @@ export function QuickAddressSelect() {
 
   return (
     <Select
-      key={currentWallet.name}
-      value={getCurrentPath(currentWallet, addresses)}
+      key={getCurrentPath(currentWallet, addresses)}
       defaultValue={getCurrentPath(currentWallet, addresses)}
       onValueChange={setCurrentAddress}
     >
--- a/gui/src/components/QuickFastModeToggle.tsx
+++ b/gui/src/components/QuickFastModeToggle.tsx
@@ -15,6 +15,7 @@ export function QuickFastModeToggle() {
     <div className="flex items-center space-x-2">
       <Switch
         defaultChecked={fastMode}
+        checked={fastMode}
         onCheckedChange={onChange}
         id="fast-mode"
       />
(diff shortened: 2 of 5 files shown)


gui/src/store/useWallets.ts
```diff
@@ -76,7 +76,7 @@ const store: StateCreator<Store> = (set, get) => ({
         ...(addresses || []).map(({ key, address }) => {
           return {
             id: `${actionId}/${wallet.name}/${key}`,
-            text: address,
+            text: `${wallet.name}/${address}`,
```
naps62, line 79:
```suggestion
            text: `${wallet.name}/${alias || address}`,
```

gui/src/store/useWallets.ts
```diff
@@ -76,7 +76,7 @@ const store: StateCreator<Store> = (set, get) => ({
         ...(addresses || []).map(({ key, address }) => {
```
naps62, line 76:
```suggestion
        ...(addresses || []).map(({ key, address, alias }) => {
```

gui/src/store/useWallets.ts
naps62: I'm assuming this could work too?

gui/src/components/QuickAddressSelect.tsx
```diff
@@ -29,6 +29,7 @@ export function QuickAddressSelect() {
   return (
     <Select
       key={currentWallet.name}
+      value={getCurrentPath(currentWallet, addresses)}
```
naps62, line 32: I don't have enough context to remember, can you explain why this is needed?
Seeing both `value` and `defaultValue` as props seems weird

gui/src/components/QuickFastModeToggle.tsx
```diff
@@ -15,6 +15,7 @@ export function QuickFastModeToggle() {
     <div className="flex items-center space-x-2">
       <Switch
         defaultChecked={fastMode}
+        checked={fastMode}
```
naps62, line 18: same as above here. `defaultXXX` is different from `XXX`, in the sense that it won't update if the value changes (on purpose). this seems suspicious

gui/src/components/QuickAddressSelect.tsx
DavideSilva: setting `value` is needed to update the displayed value when we change the address from the command bar, otherwise it wouldn't update

naps62: no, what you need is to force a re-render of the react object
setting `value` explicitly will conflict with the ability to use the input itself

possibly what you want is to set a `key={...}` to make sure it re-renders that component when some outside value change, but still have it react to updates in the select itself

DavideSilva: from what I tested, it doesn't seem to conflict with anything, I can still change the value from clicking the ui or by using the command bar

DavideSilva: ok, changing the `key` value from `currentWallet.name` to getting the current path seems to also work

naps62 (approved):

PR #979, opened 2025-02-11
Save and restore sync status for Anvil chains

### LINKED ISSUE #932: don't reset anvil state unless necessary
We're currently purging all data from anvil by default on app startup, and when we detect a rollback.
this is fine for local short-lived instances, which is what we've been doing so far

but with hosted anvil instances, such as our private testnet, it actually causes a long startup time if too much history needs to be fetched. It also adds extra load on the node itself

Instead of the current naive approach, we can do better:
- [ ] #970
- [ ] when resetting the anvil listener, instead of deleting all data right away, first check if the known hash still exists
- [ ] if hash exists, resume syncing from that block number, without deleting prior data
- [ ] if hash does not exist, do the previous behaviour (start sync from scratch)

DavideSilva:
Why:
* To fix #932

How:
* Keeping track for all anvil chains of the latest known block number
* Skipping the db reset logic when the anvil provider still knows the
  saved known block number

### DIFF
--- a/crates/sync/anvil/src/tracker.rs
+++ b/crates/sync/anvil/src/tracker.rs
@@ -132,7 +132,6 @@ async fn watch(
             .await
             .unwrap_or(serde_json::Value::Null);
 
-        // max between fork number and from_block +1
         let fork_block = node_info["forkConfig"]["forkBlockNumber"]
             .as_u64()
             .unwrap_or(0);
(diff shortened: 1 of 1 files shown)


crates/sync/anvil/src/tracker.rs
```diff
@@ -125,17 +131,21 @@ async fn watch(
             .raw_request("anvil_nodeInfo".into(), ())
             .await
             .unwrap_or(serde_json::Value::Null);
+
+        // max between fork number and from_block +1
```
DavideSilva, line 135:
```suggestion
```

naps62 (approved):

PR #983, opened 2025-02-11
update extension page url

naps62:
the site will now be a static tanstack-router page instead of next.js, so we need to move to hash-based routing


naps62:
nevermind, this is not needed

Issue #980, opened 2025-02-11
Remove ResetAnvilListener internal message


DavideSilva:

naps62:
Fixed in #986

PR #986, opened 2025-02-11
Remove ResetAnvilListener

DavideSilva:
Why:
* #980

How:
* Checking for `NetworkAdded`, `NetworkUpdated`, `NetworkRemoved`
  messages instead to trigger the reset of Anvil listeners
* Removing `ResetAnvilListener` internal message from the broadcast and
  all function calls that depended on it

### DIFF
--- a/crates/sync/anvil/Cargo.toml
+++ b/crates/sync/anvil/Cargo.toml
@@ -11,6 +11,7 @@ authors.workspace = true
 [dependencies]
 ethui-db.workspace = true
 ethui-abis.workspace = true
+ethui-networks.workspace = true
 ethui-types.workspace = true
 ethui-broadcast.workspace = true
 ethui-proxy-detect.workspace = true
--- a/crates/sync/anvil/Cargo.toml
+++ b/crates/sync/anvil/Cargo.toml
@@ -11,7 +11,6 @@ authors.workspace = true
 [dependencies]
 ethui-db.workspace = true
 ethui-abis.workspace = true
-ethui-networks.workspace = true
 ethui-types.workspace = true
 ethui-broadcast.workspace = true
 ethui-proxy-detect.workspace = true
--- a/crates/sync/anvil/src/init.rs
+++ b/crates/sync/anvil/src/init.rs
@@ -21,12 +21,11 @@ async fn receiver() -> ! {
         match rx.recv().await {
             Ok(InternalMsg::NetworkAdded(network))
             | Ok(InternalMsg::NetworkUpdated(network))
-            | Ok(InternalMsg::NetworkRemoved(network)) => {
-                if network.is_dev().await {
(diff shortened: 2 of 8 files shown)


DavideSilva:
Putting in draft until I manage to understand why this change is now causing this error on startup
```thread 'tokio-runtime-worker' panicked at crates/networks/src/init.rs:60:24:
called `Option::unwrap()` on a `None` value
``

File: crates/sync/anvil/src/init.rs [hunk 1]
```diff
@@ -18,13 +21,35 @@ async fn receiver() -> ! {
...
+            _ => (),
         }
     }
 }
 
+async fn anvil_network(chain_id: u32) -> Result<Option<(Url, Url)>> {
+    let networks = Networks::read().await;
+
+    if let Some(network) = networks.get_network(chain_id) {
+        if networks.validate_chain_id(chain_id) && network.is_dev().await {
```
naps62, line 42: if `get_network` returned something, the `validate_chain_id` is useless

crates/sync/anvil/src/init.rs
```diff
@@ -18,13 +21,35 @@ async fn receiver() -> ! {
...
+            let http = Url::parse(&network.http_url)?;
+            let ws = Url::parse(&network.ws_url())?;
+
+            return Ok(Some((http, ws)));
+        }
+    }
```
naps62, line 48: I don't like these nested ifs. can we move these to the same syntaax we used earlier today?

crates/sync/anvil/Cargo.toml
```diff
@@ -11,6 +11,7 @@ authors.workspace = true
 [dependencies]
 ethui-db.workspace = true
 ethui-abis.workspace = true
+ethui-networks.workspace = true
```
naps62, line 14: this will no longer be needed right?

crates/sync/anvil/src/init.rs
```diff
@@ -18,9 +18,17 @@ async fn receiver() -> ! {
...
 
     loop {
-        if let Ok(InternalMsg::ResetAnvilListener { chain_id, http, ws }) = rx.recv().await {
-            trace!("resetting anvil listener for chain_id {}", chain_id);
-            reset_listener(chain_id, http, ws).await
+        match rx.recv().await {
+            Ok(InternalMsg::NetworkAdded(network))
+            | Ok(InternalMsg::NetworkUpdated(network))
+            | Ok(InternalMsg::NetworkRemoved(network)) => {
+                if network.is_dev().await {
```
naps62, line 25:
```suggestion
            Ok(InternalMsg::NetworkAdded(network))
            | Ok(InternalMsg::NetworkUpdated(network))
            | Ok(InternalMsg::NetworkRemoved(network)) if network.is_dev() => {
                
```

naps62: I believe this should work too

PR #995, opened 2025-02-12
Move Network to `ethui_types`

DavideSilva:
Why:
* In order to prevent a cyclic dependency and so we can use the
  `Network` struct for sending as parameter in Network related events

How:
* Moving the `networks/src/network.rs` file to `types/src/network.rs`
* Updating references to point to the new location
* Removing `force_is_anvil` field as it is no longer used

### DIFF
--- a/crates/types/src/network.rs
+++ b/crates/types/src/network.rs
@@ -2,12 +2,10 @@ use alloy::{
     network::Ethereum,
     providers::{ext::AnvilApi, ProviderBuilder, RootProvider},
     rpc::client::ClientBuilder,
-    transports::layers::RetryBackoffLayer,
+    transports::{layers::RetryBackoffLayer, RpcError, TransportErrorKind},
 };
 use serde::{Deserialize, Serialize};
 use url::Url;
-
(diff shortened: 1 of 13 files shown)


crates/types/src/network.rs
```diff
@@ -116,13 +108,7 @@ impl Network {
...
     }
 
-    pub async fn reset_listener(&self) -> Result<()> {
-        if self.is_dev().await {
-            let http = Url::parse(&self.http_url)?;
-            let ws = Url::parse(&self.ws_url())?;
-            ethui_broadcast::reset_anvil_listener(self.chain_id, http, ws).await;
-        }
-
+    pub async fn reset_listener(&self) -> Result<(), RpcError<TransportErrorKind>> {
```
DavideSilva, line 111: This is a temporary change just to make the app compile. It's necessary to do so as moving this here adds a cyclic dependency with `ethui_broadcast`

naps62: I'm ok with it :+1:

naps62 (approved):

PR #1011, opened 2025-02-14
Network form updates

### LINKED ISSUE #982: Network form validations
- [ ] Editing a network should have different restrictions from creating a new network

- [ ] Can't use an existing network name when editing a network

- [ ] RPC should set and validate the chain id and only display the value

DavideSilva:
Why:
* Closes #982
* If we can, we should fetch the chain id from the RPC url provided when
  creating a new network
* We should also prevent creating a new, or edit, a network to one with
  an existing name

How:
* Adding a command to fetch a chain id from a provider http url
* Preventing network edits if network name exists
* Updating the network creation form to populate chain id field with
  value returned by the provider, if available

Demo of fetching chain id from http url
https://github.com/user-attachments/assets/0745ea87-debe-4f84-b07b-bd0d962ec734

### DIFF
--- a/crates/networks/src/lib.rs
+++ b/crates/networks/src/lib.rs
@@ -8,7 +8,10 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use alloy::{network::Ethereum, providers::RootProvider};
+use alloy::{
+    network::Ethereum,
+    providers::{Provider, ProviderBuilder, RootProvider},
+};
 use ethui_types::{Affinity, Network, UINotify};
 pub use init::init;
 use serde::Serialize;
@@ -82,7 +85,7 @@ impl Networks {
     pub async fn add_network(&mut self, network: Network) -> Result<()> {
         // TODO: need to ensure uniqueness by name, not chain id
         if self.validate_chain_id(network.chain_id) {
-            return Ok(());
+            return Err(Error::AlreadyExists);
         }
 
         if self.networks.contains_key(&network.name) {
@@ -98,6 +101,10 @@ impl Networks {
     }
 
     pub async fn update_network(&mut self, old_name: &str, network: Network) -> Result<()> {
+        if network.name != old_name && self.networks.contains_key(&network.name) {
+            return Err(Error::AlreadyExists);
+        }
+
         self.networks.remove(old_name);
         self.networks
             .insert(network.clone().name.clone(), network.clone());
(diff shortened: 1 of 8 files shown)


naps62:
the issue that is closed by this include "Editing a network should have different restrictions from creating a new network"

and I believe one of the details we talked about here is that you shouldn't be able to edit the chain ID, only create a new network. that is missing from this, right?

crates/networks/src/lib.rs
```diff
@@ -98,6 +104,10 @@ impl Networks {
     }
 
     pub async fn update_network(&mut self, old_name: &str, network: Network) -> Result<()> {
+        if self.networks.iter().any(|(_, n)| n.name == network.name) {
+            return Err(Error::AlreadyExists);
+        }
```
naps62, line 109: won't this always return an error if you update a network without changing its name?
it will find itself in this loop

gui/src/routes/home/_l/settings/_l/networks/_l/$name.edit.tsx
```diff
@@ -30,9 +32,49 @@ function Content({ network }: { network: Network }) {
...
+        form.setError("chain_id", {
+          type: "manual",
+          message: `RPC not online`,
+        });
+        return null;
+      }
+    };
+
+    fetchChainId();
+  }, [httpUrl, form.clearErrors, form.setValue, form.setError]);
```
naps62, line 65: these all should be part of the zod schema validations, not a use effect

crates/networks/src/lib.rs
```diff
@@ -98,6 +104,10 @@ impl Networks {
     }
 
     pub async fn update_network(&mut self, old_name: &str, network: Network) -> Result<()> {
```
naps62, line 106: I think we should have a separate struct here, which is the network without chain ID
strictly speaking it's not part of the intended arguments.

and since this is currently disabled only on the form level, it's still technically possible to change

that said, I'm ok with opening a separate issue to deal with this later

naps62 (changes requested):

crates/networks/src/lib.rs
DavideSilva: I'll make those changes together with the `internal_id` since I'll be touching that part of the code either way

crates/networks/src/lib.rs
DavideSilva: Yes, it will, nice catch

gui/src/routes/home/_l/settings/_l/networks/_l/$name.edit.tsx
```diff
@@ -46,7 +55,7 @@ function Content({ network }: { network: Network }) {
     <Form form={form} onSubmit={create} className="gap-4">
       <div className="flex flex-row gap-2">
         <Form.Text label="Name" name="name" />
-        <Form.NumberField label="Chain Id" name="chain_id" />
+        <Form.Text disabled={true} label="Chain Id" name="chain_id" />
```
naps62, line 58: should this be a NumberField again?

naps62 (approved):

PR #1046, opened 2025-02-20
Add internal_id field to Networks

DavideSilva:
Why:
* #1040

How:
* Adding a new field to the `Network` struct, `deduplication_id`
  intended to distinguish between multiple chains with the same chain
  id
* Adding a `internal_id` function to `Networks` to retrieve the tuple
  compose of chain id and deduplication id
* Adding `get_network_by_name` and `chain_id_count` to `Networks`
* Updating `ChainAdd` and `ChainUpdate` functions to take into account
  the deduplication id
* Adding a migration to handle the update to `networks.json`

### DIFF
--- a/crates/networks/src/lib.rs
+++ b/crates/networks/src/lib.rs
@@ -13,7 +13,7 @@ use alloy::{
     network::Ethereum,
     providers::{Provider, ProviderBuilder, RootProvider},
 };
-use ethui_types::{Affinity, Network, UINotify};
+use ethui_types::{Affinity, Network, NewNetworkParams, UINotify};
 pub use init::init;
 use migrations::LatestVersion;
 use serde::{Deserialize, Serialize};
@@ -95,7 +95,15 @@ impl Networks {
             .cloned()
     }
 
-    pub async fn add_network(&mut self, network: Network) -> Result<()> {
+    pub fn get_network_by_name(&self, name: &str) -> Option<Network> {
+        self.inner
+            .networks
+            .values()
+            .find(|n| n.name == name)
+            .cloned()
+    }
+
+    pub async fn add_network(&mut self, network: NewNetworkParams) -> Result<()> {
         // TODO: need to ensure uniqueness by name, not chain id
         if self.validate_chain_id(network.chain_id) {
             return Err(Error::AlreadyExists);
@@ -105,9 +113,13 @@ impl Networks {
             return Err(Error::AlreadyExists);
         }
 
+        let deduplication_id = self.get_chain_id_count(network.chain_id);
+        let network = network.into_network(deduplication_id);
+
         self.inner
             .networks
             .insert(network.name.clone(), network.clone());
+
         self.save()?;
         ethui_broadcast::network_added(network.clone()).await;
         ethui_broadcast::ui_notify(UINotify::NetworksChanged).await;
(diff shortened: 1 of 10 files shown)


crates/networks/src/lib.rs
```diff
@@ -92,6 +92,8 @@ impl Networks {
             return Err(Error::AlreadyExists);
         }
 
+        network.internal_id = Some(self.networks.len() as u32);
```
naps62, line 95: what about networks that are deserialized directly from `networks.json`? we need to find a way to set the ID for them too when initializing

crates/types/src/network.rs
```diff
@@ -21,6 +22,7 @@ pub struct Network {
 impl Network {
     pub fn mainnet() -> Self {
         Self {
+            internal_id: Some(0),
```
naps62, line 25: this isn't right. we can't assume this will have id 0. in fact it should probably have `None`, and be filled only when actually added to the persistent settings

crates/networks/src/lib.rs
DavideSilva: unless I'm missing, if the network is already in the `networks.json` that means it was added by this code and already has an `internal_id` field set

or is this to support adding a new network by manually editing the `networks.json` file?

crates/networks/src/init.rs
```diff
@@ -30,15 +30,24 @@ pub async fn init(pathbuf: PathBuf) {
...
         let reader = BufReader::new(file);
 
-        let res: PersistedNetworks = serde_json::from_reader(reader).unwrap();
+        let mut res: PersistedNetworks = serde_json::from_reader(reader).unwrap();
+
+        for (i, (_, n)) in res.networks.iter_mut().enumerate() {
+            if n.internal_id.is_none() {
+                n.internal_id = Some(i as u32)
+            }
+        }
```
naps62, line 39: this code can end up assigning duplicate IDs if you happen to have a network that already has an ID, and others that don't

which is not something that we want to happen in practice, but there could potentially be bugs that lead to that state, and we absolutely don't want to end up with the chance of bricking user configurations, since that'll be a nightmare to debug

also, adding arbitrary code in the `init` function to patch the existing structs is too clunky, I think. let's please think about this from a high level and keep the code clean

namely: this is essentially a settings migration. Previously we didn't have a `internal_id` field, now we do.
In the near future, we'll have other changes, and I don't want to accumulate random struct patches in random code paths.

What I'm saying is: I don't like this solution overall. either we think of a cleaner way to approach this, or we put a pin on it and first implement some sort of "data migration"

DavideSilva: > this code can end up assigning duplicate IDs if you happen to have a network that already has an ID, and others that don't

unless you manually added the field to the file, I don't see how you could already have an `internal_id` field before this. the case where the file exists is covered by this. the case where the file doesn't and we default the networks is also covered in this `init` function

the `broadcast_init` function now also calls `save` to persist the default networks into the file and to update the existing ones too to always include the `internal_id` field

we are also adding the field whenever we add a new network, so while I agree that we don't want to brick user's configs, I'm not sure how that could happen

DavideSilva: I'm not disagreeing this should be handled in a cleaner way as any solution we would do to essentially migrate the `networks.json` file would need to stay potentially forever as we won't know for certain when it would be safe to remove that code

Maybe that's something to tackle first then, if we are planning to include further changes to this

naps62 (approved):

PR #1071, opened 2025-02-25
Add migrations to networks

DavideSilva:
Why:
* We need a way to safely migrate between versions of the
  JSON file used to store Network related data

How:
* Adding `migrations.rs` to the `networks` crate to handle migrations of
  the `networks.json` file
* Calling the `load_and_migrate` function in the `init` of the
  `networks` crate in order to check everytime we load the file is
  there's a new version, and migrate to it


crates/networks/src/migrations.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    let persisted_networks: NetworksJson = serde_json::from_reader(reader).unwrap();
+    let file_version = persisted_networks.version;
+
+    let persisted_networks = run_migrations(persisted_networks);
+
+    if persisted_networks.version > file_version {
+        let file = File::create(path).unwrap();
+        serde_json::to_writer_pretty(file, &persisted_networks).unwrap();
+    }
```
naps62, line 41: this bit here is duplicating the same logic that already exists in `networks.save()`

I'd rather we just keep a history of previous struct versions (which right now is just a single version where the `version` field doesn't yet exist, and have the existing `Networks` struct represent the current (latest) version

the rest of the logic seems good, we just need to make it rely more on type-safety. 
what we want is more similar to the examples in this crate: https://github.com/vic1707/serde-versioning

I don't want to use the crate itself, as it doesn't seem poorly maintained, but mainly the idea of keeping an enum of historical version of the struct, where the Latest is whatever we currently use in `Networks`. the `version` field would be what we use to deserialize

the reason I don't like the non-type-safety approach here is that it relies on us remembering that `NetworksJson` needs to match whatever is on `Networks`, since you currently have two separate structs serializing into the same file. this is not safe at all

Cargo.toml
```diff
@@ -104,6 +104,8 @@ rand = "0.9"
 rstest = "0.23"
 regex = "1.11"
 anyhow = { version = "1.0", default-features = false }
+tempfile = "3.17.1"
```
naps62, line 107: this should just be in `dev-dependencies` I think, since we only use it for tests

naps62 (approved):

PR #1108, opened 2025-03-07
Refactor DedupChainId into a struct

DavideSilva:
Why:
* DedupChainId should be a struct to make it impossible to send a random
  tuple without explicit casting

How:
* Refactoring the `DedupChainId` type into a struct
* Adding `Into` implementations to convert between tuples and this type
* Converting usages of the `DedupChainId` into the struct format

### DIFF
--- a/crates/types/src/affinity.rs
+++ b/crates/types/src/affinity.rs
@@ -11,12 +11,6 @@ pub enum Affinity {
     Sticky(DedupChainId),
 }
 
-impl From<(u32, u32)> for Affinity {
-    fn from(internal_id: (u32, u32)) -> Self {
-        Affinity::Sticky(internal_id.into())
-    }
-}
-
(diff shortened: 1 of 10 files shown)


crates/types/src/affinity.rs
```diff
@@ -11,6 +11,12 @@ pub enum Affinity {
     Sticky(DedupChainId),
 }
 
+impl From<(u32, u32)> for Affinity {
+    fn from(internal_id: (u32, u32)) -> Self {
+        Affinity::Sticky(internal_id.into())
+    }
+}
```
naps62, line 18: if we now have a `DedupId` type, why do we need this? feels like we should prefer to convert tuple -> DedupId -> Affinity

DavideSilva: we don't, added this while refactoring and forgot to remove it after

PR #1111, opened 2025-03-10
Network migration to use DedupChainId

DavideSilva:
Why:
* We can use the `DedupChainId` struct to identify a network instead of
  having two separate fields in the `Network` struct

How:
* Adding a migration to update the Network struct to V3
* Updating all the references to the `chain_id` field to be a call
  instead
* Using `dedup_chain_id` in gui to access the `chain_id` field
* Renaming `internal_id` to `dedup_chain_id`

### DIFF
--- a/gui/src/routes/home/_l/settings/_l/networks/_l/new.tsx
+++ b/gui/src/routes/home/_l/settings/_l/networks/_l/new.tsx
@@ -27,7 +27,7 @@ function Content() {
   });
 
   const httpUrl = form.watch("http_url");
-  const userChainId = form.watch("chain_id");
+  const userChainId = form.watch("dedup_chain_id.chain_id");
 
   useEffect(() => {
     if (!httpUrl) return;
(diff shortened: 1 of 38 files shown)


gui/src/routes/home/_l/settings/_l/networks/_l/new.tsx
```diff
@@ -56,6 +56,7 @@ function Content() {
   const onSubmit = async (data: Network) => {
     try {
       setLoading(true);
+      data.dedup_chain_id.dedup_id = 0;
```
DavideSilva, line 59: I know this isn't ideal but this is just temporary so the app remains functional, and we still preventing networks with the same chain_id so this should never fail

Since the `Network` Zod schema is shared between the create and edit form, this meant I had to also update the `NewNetworkParams` struct to have a `dedup_chain_id`, causing this submit to fail if there's no value set to `dedup_id`

I'm now working on a separate branch to see if we can refactor out the need for the NewNetworkParams and go directly to the Network but I haven't finished that yet so for now this was the solution I came up with to move things along

naps62 (approved):

PR #1112, opened 2025-03-10
New onboarding flow

### LINKED ISSUE #897: Replace onboarding with less intrusive additional tab
The onboarding does its job, but has a few annoyances:
- for some users, only some of the steps will actually be useful. it's certainly possible for someone to see use in the wallet by only installing the extension, but not configuring custom wallets or alchemy.
- the fact that we show these settings but not others (e.g.: etherscan) makes them seem more important than they really are, and actually hides other features that could be useful
- the experience of opening the app and seeing the main screen directly is arguably much better than seeing a dumbed down UI and being forced to go through an onboarding

instead, I propose:
- a single additional tab on the sidebar for onboarding
- this tab will disapear once the user fulfills all of its tasks, or dismisses it explicitly
- the tab will have a more exhaustive list of optional things the user can do (e.g.: create a new wallet, set up a live network with alchemy, add a new contract, etc). this servers as both an onboarding step and a tutorial for the app
- the user can fulfill all of these steps in the order he wishes

naps62:
The new onboarding flow is less intrusive, and more extendable

instead of a mandatory flow on first launch, with dedicated pages and components, we show a list of optional steps you can do as an additional tab in the home bar
Each step is simply a link to the appropriate URL where that can ben done, along with a skip button.
when all steps are done/skipped, the user will no longer seem them, and the tab also disappears

This makes using the app a lot more confortable, as one can get to use the app right away without having to go through the entire onboarding
it also makes it easy for us to add steps in the future and have them show up even for existing users

Fixes #897


gui/src/routes/home/_l/settings/_l/general.tsx
```diff
@@ -54,6 +53,7 @@ function SettingsGeneral() {
     resolver: zodResolver(schema),
     defaultValues: general,
   });
+  console.log(form);
```
DavideSilva, line 56:
```suggestion
```

PR #1153, opened 2025-03-19
Fix add contract network

DavideSilva:
Why:
* The `add_contract` command was checking if the current network was a
  non-anvil chain before checking for proxy implementation, instead of
  using the network selected in the dialog

How:
* Calling `is_dev` in the Network instead of reading the value from the
  `networks_is_dev` function that checks the current network

### DIFF
--- a/bin/src/commands.rs
+++ b/bin/src/commands.rs
@@ -1,7 +1,6 @@
 use alloy::providers::Provider as _;
 use ethui_db::utils::{fetch_etherscan_abi, fetch_etherscan_contract_name};
 use ethui_db::Db;
-use ethui_networks::commands::networks_is_dev;
 use ethui_proxy_detect::ProxyType;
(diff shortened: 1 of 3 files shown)


bin/src/commands.rs
```diff
@@ -43,8 +42,9 @@ pub async fn add_contract(
 
     let code = provider.get_code_at(address).await?;
     let proxy = ethui_proxy_detect::detect_proxy(address, &provider).await?;
+    let network_is_dev = network.is_dev().await;
 
-    let (name, abi) = if networks_is_dev().await? {
```
naps62, line 47: seems like we can also delete `networks_is_dev` now. don't see any usage of it in the frontend

naps62 (approved):

bin/src/commands.rs
DavideSilva: we can, yes, I had that change in another branch

PR #1179, opened 2025-03-28
Refactor add contract to use route based form

DavideSilva:
Why:
* Instead of showing a modal to insert a contract manually,
  we want to show a form in its own page

How:
* Creating a new file based route and moving the component to it
* Disabling new release notice in development

### DIFF
--- a/gui/src/routes/home/_l.tsx
+++ b/gui/src/routes/home/_l.tsx
@@ -15,7 +15,10 @@ export const Route = createFileRoute("/home/_l")({
 
 function HomePageLayout() {
   useNoticeAlchemyKeyMissing();
-  useNoticeNewVersion();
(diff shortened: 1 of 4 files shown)


gui/src/routes/home/_l.tsx
```diff
@@ -15,7 +15,10 @@ export const Route = createFileRoute("/home/_l")({
 
 function HomePageLayout() {
   useNoticeAlchemyKeyMissing();
-  useNoticeNewVersion();
+  const isDev = import.meta.env.MODE === "development";
+  if (!isDev) {
+    useNoticeNewVersion();
+  }
```
naps62, line 21: just for cleanliness, I'd prefer to have this condition inside the `useNoticeNewVersion` instead of around it

naps62 (approved):

PR #1216, opened 2025-04-10
Add remove contract to Contracts lib

DavideSilva:
Why:
* We want to be able to remove a contract

How:
* Adding a new query and tauri command to remove an individual contract from the
  `contracts` table
* Recursing if the contract was proxied to remove its proxy
* Updating the UI to trigger the removal


bin/src/commands.rs
```diff
@@ -71,3 +73,21 @@ pub async fn add_contract(
...
     }
 }
+
+#[tauri::command]
+pub async fn remove_contract(
+    chain_id: u32,
+    address: Address,
+    db: tauri::State<'_, Db>,
+) -> AppResult<()> {
+    let has_proxy = db.get_proxy(chain_id, address).await;
```
naps62, line 83: should we disallow removing contracts from anvil networks? because those will be added again anyway once if we restart the app, could make the experience weird

DavideSilva: that would happen only if you also restarted anvil, right? for remote anvil networks, restarting the app does not re-add the contracts

naps62: yes, I meant restart anvil, not the app
essentially it seems useless to remove a contract from the ui, since it will likely come back anyway

DavideSilva: not sure I agree with disallowing anvil networks to remove contracts from the UI

a case could be made for disallowing a local anvil network to remove but then why are we having different behaviours just for one specific case? won't that be more confusing?

naps62 (approved):

PR #1222, opened 2025-04-16
Bundle libusb into MacOS app bundle

DavideSilva:
Why:
* The Ledger integration depends on having `libusb` available in the
  system. Since this is not a default library, we are bundling it
  together with the app so end users don't need to install `libusb`
  explicitly

How:
* Using the `beforeBundleCommand` Tauri hook to run the `postbuild.sh`
  to update the linking of the library


package.json
```diff
@@ -6,7 +6,9 @@
   "scripts": {
     "dev": "yarn run app:dev",
     "app:dev": "yarn tauri dev --config bin/tauri.dev.conf.json --features debug --target $(yarn run rust-target)",
-    "app:build": "yarn tauri build --target $(yarn run rust-target)",
+    "app:build": "yarn run app:build:nobundle && ./scripts/postbuild.sh && yarn run app:build:bundle",
+    "app:build:nobundle": "yarn tauri build --no-bundle --target $(yarn run rust-target)",
+    "app:build:bundle": "yarn tauri bundle --bundles app,dmg --target $(yarn run rust-target)",
```
naps62, line 11: this is not used during the release process

we need to check the release action and see how we can configure it to run the script
https://github.com/tauri-apps/tauri-action

from a quick look, the `tauriScript` option might be what we need

naps62 (changes requested):

package.json
DavideSilva: It wasn't since the tauri-action only runs the `tauri build` command in a single step. Ended up finding about the `beforeBundleCommand` hook to simply run the script between the build and bundle steps

scripts/postbuild.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
naps62, line 1:
```suggestion
#!/usr/bin/env bash
```

this is more "proper", and your version also wouldn't work in nixos

bin/tauri.conf.json
```diff
@@ -4,6 +4,7 @@
   "build": {
     "beforeDevCommand": "yarn workspace @ethui/gui dev",
     "beforeBuildCommand": "yarn workspace @ethui/gui build",
+    "beforeBundleCommand": "./scripts/postbuild.sh",
```
naps62, line 7: @DavideSilva did you validate if this works with the release action?

DavideSilva: I didn't run the action itself, no. I wasn't sure if that would mess up or trigger an actual releases. But this should work with the action since the only thing the action is doing is running `tauri build` and this change happens during that command execution, similar to the `beforeBuildCommand` that we already had

naps62 (approved):

Issue #1235, opened 2025-04-29
Add Etherscan Api Key to Onboarding


joaocosta9:
Since etherscan api key is essential for adding new contracts, maybe it would be a good idea to add to the Onboarding tab

PR #1238, opened 2025-04-30
Adds etherscan API key to onboarding steps

naps62:
Most of the changes are about making `onboarding.hidden` automatically turn false whenever we add new onboarding items. this will mean new features can get easily highlighted to existing users


naps62:
Fixes #1235

PR #1256, opened 2025-05-06
docs: added a link to the Build badge

operagxsasha:


naps62:
I don't think this link is that useful (and the markup is not correct)

PR #1257, opened 2025-05-08
Add tauri command to fetch forge ABIs

DavideSilva:
Why:
* We need a way to get the ABIs we parsed from Forge to be later used in
  the UI

How:
* Adding a `fetch_forge_abis` tauri command that asks the Kameo Forge
  actor to send the values of the `abis_by_path`
* Updating the `ForgeAbi` struct to be deserealizable and to include
  `method_identifiers`

### DIFF
--- a/crates/forge/src/actor.rs
+++ b/crates/forge/src/actor.rs
@@ -151,6 +151,20 @@ impl Message<UpdateContracts> for Worker {
     }
 }
 
+pub struct FetchAbis;
+
+impl Message<FetchAbis> for Worker {
+    type Reply = Vec<ForgeAbi>;
+
+    async fn handle(
+        &mut self,
+        _msg: FetchAbis,
+        _ctx: kameo::message::Context<'_, Self, Self::Reply>,
+    ) -> Self::Reply {
+        self.abis_by_path.clone().into_values().collect()
+    }
+}
+
 impl Actor for Worker {
     type Mailbox = BoundedMailbox<Self>;
 
--- a/crates/forge/src/commands.rs
+++ b/crates/forge/src/commands.rs
@@ -1,0 +1,14 @@
+use kameo::actor::ActorRef;
+
+use crate::{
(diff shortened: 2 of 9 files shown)


crates/forge/src/actor.rs
```diff
@@ -151,6 +151,20 @@ impl Message<UpdateContracts> for Worker {
...
+    type Reply = Vec<ForgeAbi>;
+
+    async fn handle(
+        &mut self,
+        _msg: FetchAbis,
+        _ctx: kameo::message::Context<'_, Self, Self::Reply>,
+    ) -> Self::Reply {
+        self.abis_by_path.clone().into_values().collect()
+    }
+}
```
naps62, line 166: have you tested this with a large number of ABIs?
wondering if this won't be very slow once you build up a large map

crates/forge/src/commands.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use crate::{
+    abi::ForgeAbi,
+    actor::{FetchAbis, Worker},
+    error::{Error, Result},
+};
+
+#[tauri::command]
+pub async fn fetch_forge_abis() -> Result<Vec<ForgeAbi>> {
+    let actor = ActorRef::<Worker>::lookup("forge_actor")?.ok_or(Error::ActorNotFound)?;
```
naps62, line 11: why not just `"forge"` as the actor name?

crates/forge/src/actor.rs
DavideSilva: Tested with around 80 ABIs and fetch time from the UI seemed to be ~100ms

I'm aware `clone` might have performance hits once this gets very large. I was thinking we could move the entire trace parsing logic from the UI to an actor too and that would solve it but I want to have a working prototype first before attempting that refactor

naps62: sounds like a good idea (both the moving to an actor, and the leaving it for later) :+1:

PR #1275, opened 2025-05-29
Fix linux builds

naps62:
cc/ @DavideSilva


DavideSilva:
My bad, I must have left it when I was testing different targets for bundling libusb

PR #1285, opened 2025-06-03
Fetch ABI from Etherscan if forked network

DavideSilva:
Why:
* If we are connected to a forked network and we try to add a contract
  that was deployed before the fork, we can try to find its ABI on
  Etherscan

How:
* Adding `get_forked_network` to the `Network` struct to get
  the forked network metadata
* Using the forked network chain id to get the contract's ABI from Etherscan
  before inserting

### DIFF
--- a/crates/forge/src/actor.rs
+++ b/crates/forge/src/actor.rs
@@ -459,7 +459,7 @@ mod tests {
     use super::*;
 
     #[tokio::test]
-    pub async fn find_forge_tomls() -> color_eyre::Result<()> {
+    async fn find_forge_tomls() -> color_eyre::Result<()> {
         let dir = create_fixture_directories()?;
 
         let mut actor = Worker::default();


bin/src/commands.rs
```diff
@@ -48,7 +48,16 @@ pub async fn add_contract(
...
+        let network_fork = network.get_forked_network().await;
+        match network_fork {
+            Ok(Some(fork)) => (
+                fetch_etherscan_contract_name(fork.chain_id.into(), address).await?,
+                fetch_etherscan_abi(fork.chain_id.into(), address)
+                    .await?
+                    .map(|abi| serde_json::to_string(&abi).unwrap()),
+            ),
+            _ => (None, None),
+        }
```
naps62, line 60: this seems overly simplified

what if I'm adding a contract whose ABI I happen to have locally?

all this is doing is "if it's dev and a fork, fetch from etherscan"

what we discussed was: for every new contract, we try all possible sources (first local, then etherscan)

DavideSilva: does this make sense? checking the db first for a contract with the same address? if we have a local project, forge watcher will add all the contracts to the db. if we then have a fork with the same chain id, we would find the ABI in the db already

naps62: it's not checking the db, it's checking the forge abi index we have.
the db stores contract instances (addresses, and possibly ABIs that were matched)
we can also have contracts that we added manually for which we happen to have the abi on a forge project

example #1: we have the source code for a repo that's deployed on mainnet, but not verified on etherscan. we run a fork of mainnet, and want to manually add the contracts to it, and have ethui automatically pickup the force artifacts from the source code

example #2: we don't have an etherscan api key, and we expect things to be local-first

naps62 (approved):

PR #1299, opened 2025-06-20
Basic stacks setup

ZePedroResende:


bin/Cargo.toml
```diff
@@ -58,3 +59,4 @@ default = []
 # DO NOT REMOVE!!
 custom-protocol = ["tauri/custom-protocol"]
 debug = ["tauri/devtools"]
+stacks = ["ethui-stacks"]
```
naps62, line 62: I believe the syntax now is `"dep:ethui-stacks"` to make it clear it's a dependency and not another feature flag

crates/settings/src/lib.rs
```diff
@@ -85,6 +85,11 @@ impl Settings {
             ethui_tracing::parse(&self.inner.rust_log)?;
         }
 
+        if let Some(v) = params.get("stacks") {
+            println!("Setting stacks mode to {}", v);
+            self.inner.stacks = serde_json::from_value(v.clone()).unwrap();
```
naps62, line 90: no unwrap please. should end up with a `?` so the error propagates properly

gui/src/components/AppSidebar.tsx
```diff
@@ -201,5 +201,6 @@ const settingsItems = [
   { title: "Network", url: "/home/settings/networks" },
   { title: "Foundry", url: "/home/settings/foundry" },
   { title: "Tokens", url: "/home/settings/tokens" },
+  { title: "Stacks", url: "/home/settings/stacks" },
```
naps62, line 204: missing the logic to hide this menu option if the flag is not enabled
(I think hiding the link should be enough. you can technically still navigate to it via the console, but that's fine)

crates/stacks/src/actor.rs
```diff
@@ -0,0 +1,42 @@
...
+};
+use tracing::info;
+
+#[derive(Default)]
+pub struct Worker {
+    pub stacks: bool,
+}
+
+pub enum Msg {
+    SetStacks(bool),
```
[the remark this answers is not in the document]
ZePedroResende: Yup , that feels a lot more idiomatic :thinking:

crates/settings/src/lib.rs
ZePedroResende: Fixed but just as a side note ! The others `from_value` on the same function use `unwrap`

gui/src/components/AppSidebar.tsx
ZePedroResende: On the Rust side, we have the feature to know if the feature is enabled. Is there currently a way for the TypeScript side to detect whether this flag is active? If not, should we introduce a global setting value that indicates the feature is enabled ?

crates/stacks/src/docker.rs
```diff
@@ -0,0 +1,414 @@
...
+            self.socket_path = Some(PathBuf::from("/var/run/docker.sock"));
+        }
+        Ok(self.transition())
+    }
+}
+
+impl DockerManager<DockerSocketReady> {
+    pub fn check_docker_installed(self) -> color_eyre::Result<DockerManager<DockerInstalled>> {
+        let docker_bin = self.docker_bin()?;
+        info!("Checking for Docker/Podman...");
```
naps62, line 109: feels like all of these info calls should be trace! only
or in some cases just removed

crates/stacks/src/actor.rs
```diff
@@ -0,0 +1,58 @@
...
+
+pub struct Worker {
+    pub stacks: bool,
+    pub port: u16,
+    pub config_dir: PathBuf,
+}
+
+pub enum Msg {
+    SetEnabled(bool),
+}
```
naps62, line 15: instead of this, I think I prefer this approach:

```
pub struct SetEnabled(bool);

impl Message<SetEnabled> for Worker { ...
```

that way each message gets its own handler. irrelevant for now since there's only one, but you're able to split the code better, and also have different return values for each message type

crates/stacks/src/actor.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    async fn handle(
+        &mut self,
+        msg: Msg,
+        _ctx: &mut kameo::message::Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        match msg {
+            Msg::SetEnabled(enabled) => {
+                self.stacks = enabled;
+                if enabled {
```
naps62, line 28: this should handle spam better. the app seems to be spamming repeated events (which is also a problem that I'll need to check)

but regardless, if we receive this message and `enabled == self.stacks`, then this should do nothing

bin/src/commands.rs
```diff
@@ -117,3 +117,12 @@ pub async fn remove_contract(
...
     Ok(())
 }
+
+#[tauri::command]
+pub async fn is_stacks_enabled() -> bool {
+    if cfg!(feature = "stacks") {
+        return true;
+    }
+
+    false
```
naps62, line 127:
```suggestion
    cfg!(feature = "stacks")
```

crates/stacks/src/docker_test.rs
```diff
@@ -0,0 +1,171 @@
+#[cfg(test)]
```
naps62, line 1: delete this file right?

crates/stacks/src/init.rs
```diff
@@ -0,0 +1,43 @@
...
+            && let InternalMsg::SettingsUpdated = msg
+        {
+            let settings = ethui_settings::ask(GetAll)
+                .await
+                .expect("Failed to get settings");
+
+            handle
+                .tell(Msg::SetEnabled(settings.run_local_stacks))
+                .await
+                .unwrap();
```
naps62, line 40: no unwraps or expects pls

naps62 (approved):

PR #1320, opened 2025-07-02
Forge test cli subcommand

DavideSilva:
Why:
* We want to be able to capture the output from the `forge test -vvvvv
  --json` command to send to the UI so we can display the trace output
  in the UI

How:
* Adding a new command and subcommand to run `forge test -vvvvv --json`
* Adding a new websocket RPC endpoint to receive the forge test output
* Capturing the output and sending to the websocket of ethui


bin/Cargo.toml
```diff
@@ -55,6 +56,7 @@ ethui-token-list.workspace = true
 
 [features]
 default = []
+forge-traces = ["ethui-args/forge-traces", "ethui-forge-traces/forge-traces", "ethui-rpc/forge-traces"]
```
naps62, line 59: why does the forge-traces crate have a forge-traces feature flag?
you can just mark that crate as `optional = true`, and add `dep:ethui-forge-traces` here to include it as dependency when the feature is enabled

the crate itself doesn't need to know about the feature flag

bin/src/lib.rs
```diff
@@ -15,11 +22,17 @@ static LOCK_NAME: &str = "iron-wallet";
...
 
 #[tokio::main]
-pub async fn run() -> color_eyre::Result<()> {
+pub async fn run() -> Result<()> {
     ethui_tracing::init()?;
     fix_path_env::fix()?;
 
     let args = ethui_args::parse();
+
+    #[cfg(feature = "forge-traces")]
```
naps62, line 31: all these cfg usages in this file are a bit of an anti-pattern.

instead, you should:

- Add a `App` subcommand to the enum, and make that the default one. meaning that `./ethui` and `./ethui app` are the same thing
- the code you're skipping on `run` just becomes what you run when in the `App` subcommand
- the `Forge` subcommand just gets delegated to the forge traces crate (for now at least). so you can move a lot of this code inside it

.mise.toml
```diff
@@ -29,7 +29,7 @@ rust = "nightly"
 [env]
 NODE_ENV = "development"
 RUST_LOG = "warn"
-ETHUI_FEATURES = "debug"
+ETHUI_FEATURES = "debug forge-traces"
```
naps62, line 32: should be a comma here, otherwise may cause problems in some usages

bin/src/app.rs
```diff
@@ -232,7 +232,9 @@ fn config_dir(app: &tauri::App, args: &Args) -> PathBuf {
 }
 
 async fn should_start_main_window(args: &Args) -> bool {
-    if args.hidden {
+    if let Some(Commands::App { hidden }) = &args.command
+        && *hidden
+    {
```
naps62, line 237: here, I think we should just receive the `Commands::App` element, because it should be always true that the command itself is that one

so the argument of this function should change, not this condition

bin/src/lib.rs
```diff
@@ -15,22 +17,38 @@ static LOCK_NAME: &str = "iron-wallet";
...
+    }
 
     app::EthUIApp::build(&args).await?.run();
 
     Ok(())
 }
+
+async fn handle_command(command: &Commands, args: &ethui_args::Args) -> Result<()> {
+    match command {
+        Commands::App { hidden: _ } => {
```
naps62, line 37:
```suggestion
        Commands::App {  } @ app_args => {
```

use these args as the argument instead of the global ones

crates/args/src/lib.rs
```diff
@@ -13,6 +13,29 @@ pub struct Args {
...
+    Forge {
+        #[command(subcommand)]
+        subcommand: ForgeCommands,
+    },
+}
+
+#[derive(Subcommand, Debug)]
+pub enum ForgeCommands {
+    /// Run tests
+    Test,
```
naps62, line 40: Ideally, we should be able to receive arbitrary arguments here, and forward them all to forge test

e.g.: `ethui forge test --match-contract ...`
we'd only need to enforce adding `-vvvvv --json` which are the ones we do need

.mise.toml
```diff
@@ -29,7 +29,7 @@ rust = "nightly"
 [env]
 NODE_ENV = "development"
 RUST_LOG = "warn"
-ETHUI_FEATURES = "debug"
```
naps62, line 32: this can't go on the final merge

bin/src/app.rs
```diff
@@ -111,7 +111,12 @@ impl EthUIApp {
 
         init(&app, args).await?;
 
-        if should_start_main_window(args).await {
+        let hidden = if let Some(Commands::App { hidden }) = args.command {
+            hidden
+        } else {
+            false
+        };
```
naps62, line 118: `args.command.then(|c| c.hidden).unwrap_or_default()`

bin/src/app.rs
```diff
@@ -236,8 +241,8 @@ fn config_dir(app: &tauri::App, args: &Args) -> PathBuf {
         })
 }
 
-async fn should_start_main_window(args: &Args) -> bool {
-    if args.hidden {
+async fn should_start_main_window(hidden: bool) -> bool {
```
naps62, line 244: this function seems redundant now. why not just inline it?

PR #1333, opened 2025-07-11
WIP: Refactor tracker watch function

ZePedroResende:
Currently the watch function flow is complex .

By extracting some of the logic to specific function we can remove the use of labels that atleast for me , makes it a little more readable and easier to understand the flow of each of the matches and breaks .


ZePedroResende:
Changed to WIP , as i noticed this has a bug .

naps62 (approved):

PR #1334, opened 2025-07-15
Remove mise-en-place

naps62:
I have stopped using mise in other projects a while ago (package management isn't great, the node auto installation feature never worked on NixOS)
lately I've been using it only for env variables, which turns out, most of them can be removed here:
- ETHUI_FEATURES is no longer needed by default. tauri uses `debug_assertions` now to toggle inspector, so we don't need the custom `debug` feature
- default ports for dev vs prod can be handled in code

my understanding is other people aren't using this either, so should be safe to remove

@ZePedroResende @DavideSilva ?


ZePedroResende:
LGTM and it's ok on my side

DavideSilva:
I am using it but I don't have any strong feelings about keeping it in the codebase. Just one question, how are we then enforcing Node versions?

ZePedroResende:
@DavideSilva this doesn't remove the .tool-version file, so you can still use mise or asdf to manage that !

naps62:
@DavideSilva there's already a `.tool-versions` (which is what mise also uses anyway)
but we don't really need to be too strict about it I think

I don't think we've ever had bugs that were caused by different node versions

PR #1399, opened 2025-08-07
Explorer

joaocosta9:
Explorer base branch


crates/rpc/src/utils.rs
```diff
@@ -0,0 +1,28 @@
+use alloy::providers::Provider as _;
+use ethui_types::prelude::*;
+
+pub async fn get_code(address: Address, chain_id: u32) -> jsonrpc_core::Result<Option<String>> {
```
naps62, line 4: this doesn't need to returna `jsonrpc_core::Result`. can be a `color_eyre::Result` which is our default

also, instead of `String`, it should return `Bytes`

crates/rpc/src/utils.rs
```diff
@@ -0,0 +1,28 @@
...
+    
+    let params = jsonrpc_core::Params::Array(vec![
+        json!(address),
+        json!("latest"),
+    ]);
+    
+    let result = provider
+        .raw_request::<_, String>("eth_getCode".into(), params)
+        .await
+        .map_err(crate::error::alloy_to_jsonrpc_error)?;
```
naps62, line 21: I believe you can just do `provider.get_code_at(address).await?`

no need for manually coding/encoding things (it's what rust is already good at)

joaocosta9: Yup much simpler!

PR #1409, opened 2025-08-21
Explorer transactions

joaocosta9:


crates/rpc/src/lib.rs
```diff
@@ -266,6 +266,32 @@ impl Handler {
         Ok(sender.finish().await?)
     }
 
+    pub async fn get_code(address: Address, chain_id: u32) -> jsonrpc_core::Result<Option<String>> {
```
naps62, line 269: this function isn't well placed here. this handler is for implementing the JSON-RPC server from ethereum. you're wrapping a rpc endpoint but via a custom function that's not compatible with the server itself

easiest fix is to just move this function to a utils.rs module

gui/src/components/AppSidebar.tsx
```diff
@@ -86,13 +86,13 @@ export function AppSidebar() {
...
                 <CustomSidebarMenuItem key={item.title} {...item} />
               ))}
-
-              <CollapsibleMenuSection
-                icon={<Globe />}
-                title="Explorer"
-                items={explorerItems}
-              />
-
+              {isDev && (
```
naps62, line 89: why just on dev?

joaocosta9: I got the idea it was what we discussed in the last meeting. Should it be only on chain Id 31337?

naps62: oh, sorry. that's correct, I thought this meant the `NODE_ENV` kind of dev

PR #1432, opened 2025-09-05
General updates (description)

joaocosta9:
This PR

- Refactor sidebar tabs:
      - Removed the `Transactions` tab.
      - Moved the `Contract` tab into the `Explorer` tab.
- Added a `View Transactions` button to the `Account` page (see img1).
- Most addresses and hashes are now linked to the explorer.
- The `Explorer` tab is only shown when connected to the Anvil network (not when running locally).
- Added an `EmptyState` component for `Connections` (and for use in other sections in the future), since empty screens aren’t ideal (img2)

Screenshot 2025-09-05 at 18 00 19
Screenshot 2025-09-05 at 18 17 10


gui/src/routes/dialog/_l/erc721-add.$id.tsx
```diff
@@ -37,7 +37,9 @@ function ERC721AddDialog() {
         <div>
           <Datapoint
             label="Contract Address"
-            value={<AddressView address={token.contract.address} />}
+            value={
+              <AddressView showLinkExplorer address={token.contract.address} />
```
naps62, line 41: why not make `showLinkExplorer` the default instead?
I would imagine not showing the link is the exception

joaocosta9: Yup that makes sense

PR #1445, opened 2025-09-23
Analytics

### LINKED ISSUE #1443: Add some more tracking to gauge usage
First of all: we want to remain fully private. this is not meant as a tracking tool, just as a way for us to get anonymous data that features are being used

we currently have a single aptabase event for when the app is launched. this was useful as a first hello world type of integration, but we need a bit more:
- create a random unique ID (store it in the config directory for persistence) and use that on every future aptabase event (to be able to gauge unique users)
- create a few more tracking events, to help us better understand whether people are just launching the app or aactually using it.

As a first suggestion for what events to track:
- browser tab connected: now that only web3 tabs actually connect, this will be a good indicator of people having the extension and app both installed and running
- wallet connected: tag that a wallet was connected, and ideally indicate the wallet type as well (*but absolutely not the address*)
- dialog opened (along with the dialog's name)
- transaction submitted (along with the target chain ID)

joaocosta9:
Closes #1443

### DIFF
--- a/crates/analytics/src/init.rs
+++ b/crates/analytics/src/init.rs
@@ -1,0 +1,61 @@
+use ethui_broadcast::{InternalMsg, UIMsg};
+use tauri::AppHandle;
+
+use crate::track_event;
+
+pub async fn init(handle: &AppHandle) {
+    let handle = handle.clone();
+    tauri::async_runtime::spawn(async move {
+        receiver(handle).await;
+    });
+}
+
+async fn receiver(handle: AppHandle) -> ! {
+    let mut internal_rx = ethui_broadcast::subscribe_internal().await;
+    let mut ui_rx = ethui_broadcast::subscribe_ui().await;
+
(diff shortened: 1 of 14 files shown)


crates/analytics/src/init.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    loop {
+        tokio::select! {
+            internal_msg = internal_rx.recv() => {
+                if let Ok(msg) = internal_msg {
+                    use InternalMsg::*;
+                    
+                    match msg {
+                        PeerAdded => {
+                            let _ = track_event(&handle, "extension_connected", None);
```
naps62, line 30:
```suggestion
                            let _ = track_event(&handle, "peer_added", None);
```

this will actually emit events anytime we connect a new tab, so it's better to use a 1:1 mapping, since "extension_connected" could be interpreted in other ways

PR #1581, opened 2026-07-02
Adds WalletConnect support

JoelGama:
Adds WalletConnect v2 pairing support to ethui — pair with a dApp via a `wc:` URI, approve/reject the session in a dialog, forward the dApp's JSON-RPC requests through ethui's existing RPC handler, and manage sessions (disconnect, switch account) from the connections tab.

# What's included

- New `ethui-walletconnect` crate: parses the `wc:` pairing URI, connects to the relay (`relay_client/relay_rpc`, pinned to `reown-com/reown-rust`), decrypts inbound proposals (ChaCha20-Poly1305 envelopes per `@walletconnect/utils` ≥2.18), and drives the full propose → approve → settle → request-forwarding → disconnect lifecycle.
- Method capability filtering: a dApp's requested methods are checked against what `ethui_rpc::Handler` actually implements before ever being approved — an unsupported required method rejects the proposal outright with a proper error; unsupported optional methods are just dropped. Enforced again at request time, alongside session expiry.
- Session management UI: connections tab lists active sessions, lets you disconnect or switch which account is exposed to a dApp without re-pairing, refetches live via a new event instead of only on tab switch.
Hardcoded WalletConnect Cloud project ID, shipped identically in every release.
- Clean shutdown: quitting ethui disconnects every active session instead of just vanishing (closing to the tray does not disconnect).
- `cargo-deny` fixes for the new dependency tree (git-source allowlist, `ring` license clarification + unmaintained-advisory ignore).


bin/src/app.rs
```diff
@@ -128,6 +128,9 @@ impl EthUIApp {
                 ethui_sync::commands::sync_get_native_balance,
                 ethui_simulator::commands::simulator_run,
                 ethui_simulator::commands::simulator_get_call_count,
+                ethui_walletconnect::commands::wc_pair,
+                ethui_walletconnect::commands::wc_disconnect,
+                ethui_walletconnect::commands::wc_list_sessions,
```
naps62, line 133: kind of a nitpick, but they're called reown now, so maybe we should rename the crate?

it's also shorter to type, which is convenient

crates/walletconnect/src/client.rs
```diff
@@ -0,0 +1,67 @@
...
+const DEFAULT_PROJECT_ID: &str = "25df5d97cf325d46df6b1abf570a655d";
+
+/// Builds a relay `Client`, connects to the relay, and returns a channel receiver
+/// for all inbound messages.
+///
+/// Uses the `WALLETCONNECT_PROJECT_ID` env var if set, otherwise falls back to
+/// the built-in project ID.
+pub async fn connect() -> WcResult<(Client, mpsc::UnboundedReceiver<PublishedMessage>)> {
+    let project_id =
+        std::env::var("WALLETCONNECT_PROJECT_ID").unwrap_or_else(|_| DEFAULT_PROJECT_ID.to_owned());
```
naps62: this needs to either:
a) be a hardcoded thing we ship in all releases (everyone uses the same one, not sure if it's allowed)
b) we need people to bring their own key, in which case it should be into settings, not as an env var. check `ethui_settings` for more info

crates/walletconnect/src/crypto.rs
```diff
@@ -0,0 +1,122 @@
+use chacha20poly1305::{ChaCha20Poly1305, Key, KeyInit, Nonce, aead::Aead};
```
naps62: we already have a `ethui_crypto` crate with similar logic and some overlapping crates (currently only used to encrypt/decrypt passwords and secure mnemonics)

perhaps would make sense to move these functions there and use as dependency here?

crates/walletconnect/src/client.rs
JoelGama: From what I know it should be the same for everyone. 
I just have a question: how and where should it be hardcoded?

bin/src/app.rs
JoelGama: Reown is the company/SDK's new name, but the protocol itself is still called 'WalletConnect' everywhere. Renaming to `reown` would match the new branding, but it would make the code less clear about which protocol it implements.

crates/walletconnect/src/crypto.rs
JoelGama: They look similar because both touch encryption, but they're solving different problems.

`ethui_crypto` protects secrets with a password. The user types a password, and that gets turned into an encryption key.

`walletconnect/crypto.rs` encrypts messages sent between ethui and other apps over WalletConnect. There's no password involved. The encryption key comes from a key-exchange process, and the message format has to match WalletConnect's spec exactly, or other apps won't understand it.

They both happen to use the same underlying encryption algorithm (ChaCha20-Poly1305), but with different settings, and that's really the only thing they share. Merging them would either drag WalletConnect-specific logic into a crate that's meant to be simple and password-focused, or force us to invent a shared layer that's too thin to actually save any code.

What I'd suggest instead: pull out just the raw "encrypt these bytes with this key" function (no password logic, no message-format rules) into a small shared helper both crates can use.

naps62 (approved):

PR #1590, opened 2026-07-25
Skip chain add dialog in fast mode

DavideSilva:
## Summary
- skip wallet_addEthereumChain approval dialog for dev networks when fast mode is enabled
- apply the same chain-add skip path inside wallet_updateEthereumChain

## Validation
- cargo fmt
- cargo check -p ethui-rpc


naps62:
This is dangerous. it could allow 3rd party websites to silently add malicious chain RPCs
at best this should only work after verifying that the target rpc is an anvil/hardhat network

DavideSilva:
This was done during the ETHGlobal hackathon as a way to speed up the demo. Our idea was to only allow this in fast mode and for anvil chains but totally fine with closing it if it's not worth it

PR #1591, opened 2026-07-29
fix(wallets): reject impersonator/real-key address collisions

JoelGama:
## Summary

Adding a wallet whose address already belongs to another wallet is currently unguarded (`ensure_no_duplicates_of` only checks names). When the colliding pair is a real-key wallet (mnemonic/HD/private key) and an `Impersonator`, `Wallets::find` has no tie-breaker and returns whichever comes first in the list — so an impersonator can silently shadow a real wallet's address. Since `Impersonator::build_signer` always errors, any transaction from that address then fails, even though the real key is right there.

## Fix

`create`/`update` now call `ensure_no_impersonator_collision`, which rejects the new/updated wallet if it shares an address with an existing wallet of the opposite signing kind. Same-kind collisions (e.g. the same mnemonic imported twice, or two impersonators) are unaffected and stay allowed.

### DIFF
--- a/crates/wallets/src/lib.rs
+++ b/crates/wallets/src/lib.rs
@@ -115,7 +115,7 @@ impl Wallets {
         let addresses = wallet.get_all_addresses().await;
 
         self.ensure_no_duplicates_of(&wallet.name())?;
-        self.ensure_no_impersonator_collision(&wallet, None).await?;
+        self.ensure_no_cross_tier_collision(&wallet, None).await?;
 
         self.wallets.push(wallet);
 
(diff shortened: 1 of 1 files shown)


naps62:
@JoelGama this is nice, I never thought of this backdoor

one question though: shouldn't we also do the same check against plaintext wallets? i.e. if I try to add an insecure private key for something that already exists as a secure wallet

JoelGama:
@naps62 
Actually, this issue was reported to me by @ZePedroResende, and I only implemented the fix.

Regarding the concern you raised, it can indeed happen. I suggest introducing a `SecurityTier` enum with the following hierarchy:

```rust
None (Impersonator) < Unencrypted (Plaintext) < Protected (HD / PrivateKey / JsonKeystore / Ledger)
```

Then, add a new method, `ensure_no_cross_tier_collision`, which rejects a wallet if its address collides with an existing wallet belonging to a different security tier.

Same-tier collisions should remain allowed. For example:

- Two impersonator wallets with the same address.
- The same mnemonic imported twice.
- An HD wallet and a Ledger wallet deriving the same address.

This validation should be wired into both wallet creation and updates:

- **Create:** perform the check before inserting the new wallet.
- **Update:** perform the check before replacing the existing wallet, excluding the wallet's own previous entry (identified by name) from the collision check.

